PR #68, opened 2024-10-02
Expose two customizations for `TextField` on Android

boctor:
* Specify a `VisualTransformation` for changing the visual output of the text
* Specify an `onValueChange` closure to process the textfield's text as it changes

If these customizations aren't provided, then the previous behavior is maintained:
* For `VisualTransformation`, `PasswordVisualTransformation` is used when `isSecure` otherwise `VisualTransformation.None` is used
* For `onValueChange`, the text binding is set to the new text value without change

Example usage:
```
    let textField = TextField("My TextField", text: $text)
    #if SKIP
    textField.visualTransformation = MyCustomVisualTransformation()
    textField.onValueChange = { newValue in
      text.wrappedValue = updateText(newValue)
    }
    #endif
```

Skip Pull Request Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,20 @@ With `ComposeView` and the `Compose()` function, you can move fluidly between Sw
 
 `ComposeView` and the `Compose()` function are only available in Android, so you must guard all uses with the `#if SKIP` or `#if os(Android)` compiler directives. 
 
+### Additional Considerations
+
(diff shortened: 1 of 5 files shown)


aabewhite:
Thanks! I'm curious about the motivation here. What are you using the visual transformation for, and how are you accomplishing the same on the iOS side? And why the onChange block rather than using SwiftUI's own onChange modifier?

I think we're close to landing on a more comprehensive way of doing Android-specific component customization. So I'll be honest that I don't think we'll use this patch, but seeing what you need and understanding the motivations is definitely helping shape the API we're coming up with.

PR #83, opened 2024-10-30
Made ComposeModifierView public.

moros:

Would be useful to make `ComposeModifierView` public to allow a 3rd party codebase to define their own view extensions. I was exploring creating my own extension as mentioned on the documentation but couldn't because of type visibility. The other types, like ComposeContext and ComposeResult are already public. I ran tests though I see existing failing cases; an @testable import makes internal types visible to the unit test meaning increasing the visibility more wouldn't contribute to test failers.

Skip Pull Request Checklist:

- [ ] OPTIONAL: I have tested my change on an Android emulator or device

### DIFF
--- a/Sources/SkipUI/SkipUI/Compose/ComposeModifierView.swift
+++ b/Sources/SkipUI/SkipUI/Compose/ComposeModifierView.swift
@@ -23,14 +23,14 @@ public enum ComposeModifierRole {
 }
 
 /// Used internally by modifiers to apply changes to the context supplied to modified views.
-class ComposeModifierView: View {
+public class ComposeModifierView: View {
     let view: View
     let role: ComposeModifierRole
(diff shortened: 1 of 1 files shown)


aabewhite:
Thank you.

Making an additional type public is a commitment of sorts that we'll try to not to make breaking changes to its API. So we'd prefer to minimize the public surface area of the module. Can you provide an example of something that can't be done via the existing Compose integration or a custom SwiftUI modifier?

moros:
The documentation under the Modifiers section seems to imply that its available for use outside of the skip-ui module. The `composeModifier` as mentioned in the docs seems like it'll work for what I need currently. If I wanted to create my own SwiftUI view type and extend it accessing the context and view kind of like in the public onSubmit function in TextField then I would need access to the ComposeModifierView. Though this would probably be more of a rare occurrence.

PR #94, opened 2024-12-26
Add ProgressView labels

ky-is:
Displays ProgressView labels according to their style matching SwiftUI.

Additional things to review:
- I used `let label: (any View)?` instead of the reference impl `let label: ComposeBuilder` because it made this cleaner. I don't know if there's a meaningful difference, but happy to switch if preferred.
- In theory there should be another set of initializers for setting `total` to nil with a label, but when `value` is nil, `total` is always ignored in current codepaths anyway.
- There weren't existing ProgressView tests to add to, but I made a quick UI validation screen separately. Not sure if this is worth saving somewhere here.
---

### DIFF
--- a/Sources/SkipUI/SkipUI/Components/ProgressView.swift
+++ b/Sources/SkipUI/SkipUI/Components/ProgressView.swift
@@ -69,19 +69,21 @@ public struct ProgressView : View {
         }
         if style == .linear {
             if let label, !EnvironmentValues.shared._labelsHidden {
+                let contentContext = context.content()
                 Column(modifier: context.modifier, verticalArrangement: Arrangement.spacedBy(3.dp), horizontalAlignment: androidx.compose.ui.Alignment.Start) {
-                    label.Compose(context: context.content())
-                    ComposeLinearProgress(context: context)
+                    label.Compose(context: contentContext)
+                    ComposeLinearProgress(context: contentContext)
                 }
             } else {
                 ComposeLinearProgress(context: context)
             }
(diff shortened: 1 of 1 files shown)


marcprux:
> * There weren't existing ProgressView tests to add to, but I made a quick UI validation screen separately. Not sure if this is worth saving somewhere here.

The Showcase [ProgressViewPlayground.swift](https://github.com/skiptools/skipapp-showcase/blob/main/Sources/Showcase/ProgressViewPlayground.swift) would be a good place to put it.

aabewhite (changes requested):
Thanks for the contribution! Just a couple things:

- Please do use ComposeBuilder. Having a consistent approach makes it easier to do large-scale changes across the module later if needed. I think if you just update your label constructors to delegate to the version that takes a @ViewBuilder and use ComposeBuilder.from() within that constructor, it won't make the code messy
- The context.modifiers should only be applied once on the top-level Composable. So if render a Column, the Column should apply context.modifiers, but the Composables within the Column should not. Otherwise if for example the modifiers contain a .border, you'll be displaying a border around the entire component plus another border around each element. I generally create a "contentContext = context.content()" and use that for all the content within my outermost container

ky-is:
Thank you Abe!
- Refactored to ComposeBuilder. Toggle could potentially use the same update.
- Gotcha, still figuring out context. Hopefully this is correct now...

aabewhite (approved):
Looks great!

PR #95, opened 2024-12-27
Add page TabView style

ky-is:
First attempt at supporting the page scrolling TabView style. There's still details I don't totally understand. Putting this up now to get confirmation if this approach is acceptable or not.

I added examples to `TabViewPlayground` on this branch: https://github.com/ky-is/skipapp-showcase/tree/tabview-page-style

Additional things for review:
- Recomposes: I tried to audit this and I think I've minimized them but it should be confirmed I'm using `remember` correctly (especially after any additional changes).
- I'm not sure on exactly how to handle the layout for the outer container. This version might crash in certain configurations if it can't figure out how to size itself (I don't currently have any repros).
- Page indicator UI: I just tried to make it look like SwiftUI since there isn't a standard Android indicator overlay I saw.
- I'm applying `aspectRatio` because I want it for my own use case. But currently Skip considers `aspectRatio` to be an `Image`-only modifier. It should probably be abstracted to apply to other views (and respect `contentMode` here...). I'm just not sure the best way to go about this.
- There is also `IndexViewStyle`, it should probably be implemented or stubbed along side this.

---

### DIFF
--- a/Sources/SkipUI/SkipUI/Containers/TabView.swift
+++ b/Sources/SkipUI/SkipUI/Containers/TabView.swift
@@ -11,9 +11,11 @@ import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.aspectRatio
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.ExperimentalLayoutApi
+import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.asPaddingValues
 import androidx.compose.foundation.layout.fillMaxSize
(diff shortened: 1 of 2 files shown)


aabewhite (commented):
I'll get to some specific code review comments later, but generally looks good. My main question is whether you've tested this in situations where the TabView does not have a constrained .frame(height:) or .aspectRatio?

ky-is:
> My main question is whether you've tested this in situations where the TabView does not have a constrained .frame(height:) or .aspectRatio?

I added some more examples, I didn't have a good illustration of that. By default the page style on iOS expands to all available space (when outside ScrollViews). This does the same on Android.

The collapsing behavior inside a scrollview is slightly different though. Notably, an unconstrained TabView on Android will resize to fit the page you swipe to, where on iOS it is constant based on the first page.

aabewhite:
OK, cool. I'm in the middle of some other work and my hours during the holidays have been spotty, but I'll try to merge this in the next couple of days. It's in such good shape that I don't think we have to iterate in code review. Instead I think I'll play with it in main and add any tweaks I find there, then I'll tag it for release when I'm convinced it's ready. Thanks again!

ky-is:
Sounds great, enjoy the holidays Abe!

I found this crash calculating layout:
```swift
ScrollView {
    VStack {
        TabView {
            Rectangle()
        }
        .tabViewStyle(.page)
    }
}
java.lang.IllegalStateException: Asking for intrinsic measurements of SubcomposeLayout layouts is not supported. This includes components that are built on top of SubcomposeLayout, such as lazy lists, BoxWithConstraints, TabRow, etc. To mitigate this:
- if intrinsic measurements are used to achieve 'match parent' sizing, consider replacing the parent of the component with a custom layout which controls the order in which children are measured, making intrinsic measurement not needed
- adding a size modifier to the component, in order to fast return the queried intrinsic measurement.
```

- I played with `aspectRatio` more and this version produces surprisingly consistent results between platforms.
- `IndexViewStyle` seems so niche that it's not worth it.
- On iOS if you specify `.tabItem` with an image in the first child, the page indicator dot is replaced with that image. Neat for putting like the search icon on the old iOS homescreen but also niche.

PR #96, opened 2024-12-27
Fix blank navigation title space when title is unset

ky-is:
When NavigationStack children don't have a title set, SkipUI leaves blank space where the title would be, inconsistent with SwiftUI. This can currently be worked around by explicitly applying `.navigationBarTitleDisplayMode(.inline)`.

Minimal repro:
```swift
struct ContentView: View {
    var body: some View {
        NavigationStack {
            ScrollView {
                NavigationLink("Root") {
                    ScrollView {
                        Text("Subview")
                    }
                }
            }
        }
    }
}
```
On iOS:
 

Image 1 is the same before and after, image 3 is with the fix:
  

---


aabewhite (approved):
Makes sense and changes LGTM

PR #105, opened 2025-01-17
Add FocusState and focused modifiers

ky-is:
Proof of concept for controlling focus of TextFields. Not sure this is the best way to do it but it would be great to have support for this!

Caveats:
- Seems like we would need transpiler changes to support the property wrapper syntax
- There isn't TextField-specific code here but I haven't tested it with anything else yet. In theory any focusable composable is supported https://developer.android.com/develop/ui/compose/touch-input/focus/change-focus-behavior#making-composable-focusable
- Should consider adding `focusable`/`defaultFocus`/etc modifiers as well
- There could be a better approach than every `focused` modifier remembering its own `FocusRequester`. This can create weird behavior if you try (like assigning `.focused(focus)` to multiple fields creates an infinite loop), but it seems fine for typical usage
- Need to figure out how the initializer can determine the type of `Value` so `_wrappedValue` can default to `nil` or `false` since `init() where Value: Bool` isn't supported

Example of the current usage without property wrapper support:
```swift
struct ContentView: View {
    @State var textA = ""
    @State var textB = ""

    #if !SKIP
    @FocusState var focusBool: Bool
    @FocusState var focusHashable: String?
    #else
    var focusBool = FocusState<Bool>()
    var focusHashable = FocusState<String?>()
    #endif

    var body: some View {
        Form {
            #if !SKIP
            Text("Boolean: \(focusBool.description)")
            Text("Hashable: \(focusHashable?.description ?? "nil")")
            TextField("textA", text: $textA)
               .focused($focusBool)
               .focused($focusHashable, equals: "textA")
            TextField("textB", text: $textB)
                .focused($focusHashable, equals: "textB")
            #else
            Text("Boolean: \(focusBool.wrappedValue.description)")
            Text("Hashable: \(focusHashable.wrappedValue)")
            TextField("textA", text: $textA)
                .focused(focusBool)
                .focused(focusHashable, equals: "textA")
            TextField("textB", text: $textB)
                .focused(focusHashable, equals: "textB")
            #endif
        }
        .onAppear {
            #if !SKIP
            focusHashable = "textB"
            #else
            focusHashable.wrappedValue = "textB"
            #endif
        }
    }
}
```

---

### DIFF
--- a/Sources/SkipUI/SkipUI/System/Focus.swift
+++ b/Sources/SkipUI/SkipUI/System/Focus.swift
@@ -4,6 +4,72 @@
 // under the terms of the GNU Lesser General Public License 3.0
 // as published by the Free Software Foundation https://fsf.org
 
+#if SKIP
+import SkipModel
+
+import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.SideEffect
+import androidx.compose.ui.focus.FocusRequester
+
+public final class FocusState<Value>: StateTracker {
+    public init() {
+        _wrappedValue = initialValue
+        StateTracking.register(self)
+    }
+
+    internal var initialValue: Value {
+        return false as Value // FIXME: Should be set to `nil` or `false` based on the type of Value, but it must be reified to check?
+    }
+
+    public var wrappedValue: Value {
+        get {
+            if let _wrappedValueState {
+                return _wrappedValueState.value
+            }
+            return _wrappedValue
+        }
+        set {
+            _wrappedValue = newValue
+            _wrappedValueState?.value = _wrappedValue
+        }
+    }
+    private var _wrappedValue: Value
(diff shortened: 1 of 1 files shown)


aabewhite:
Fantastic! FocusState has been moving up our priority list, so it's amazing to have all the basics here. Hopefully I can start working with patch over the weekend, because we do indeed need to add transpiler support alongside it. Thanks!

aabewhite (approved):
I'm going to merge this in - it can't hurt anything - and iterate from here. Thanks!

marcprux:
This is a fantastic addition. Thanks again for this, @ky-is!

I just added this to the Skip Notes sample ([ContentView.swift](https://github.com/skiptools/skipapp-notes/blob/main/Sources/SkipNotes/ContentView.swift#L125)), and it works great.

Screenshot 2025-01-21 at 15 38 01

ky-is:
Awesome yea definitely a nice qol thing!

PR #123, opened 2025-02-11
Render Section header/footer outside of List

ky-is:
Section has special behavior inside SwiftUI List. But otherwise, it simply renders the optional header/footer before/after the content you pass to it. This attempts to match that behavior.

---

### DIFF
--- a/Sources/SkipUI/SkipUI/Containers/Section.swift
+++ b/Sources/SkipUI/SkipUI/Containers/Section.swift
@@ -70,7 +70,13 @@ public struct Section : View, LazyItemFactory {
 
     #if SKIP
     @Composable override func ComposeContent(context: ComposeContext) {
+        if let header {
+            header.Compose(context: context)
+        }
         content.Compose(context: context)
+        if let footer {
+            footer.Compose(context: context)
+        }
     }
 
     @Composable func appendLazyItemViews(to composer: LazyItemCollectingComposer, appendingContext: ComposeContext) -> ComposeResult {


Sources/SkipUI/SkipUI/Containers/Section.swift
```diff
@@ -70,7 +70,13 @@ public struct Section : View, LazyItemFactory {
 
     #if SKIP
     @Composable override func ComposeContent(context: ComposeContext) {
+        if let header {
```
aabewhite, line 73: I'm not familiar with SwiftUI's behavior here: should this be in its own Column or other container? In addition to layout, I worry that e.g. the modifier chain is going to be applied to the header, footer, and content separately.

ky-is: Yeah this produces incorrect results when a Section is not nested inside a layout container (though semantically it doesn't necessarily make sense outside one anyway). iOS seems to fall back to a centered VStack, but only if there isn't an existing layout. Is there a way to access the current layout context so we know if we're in an HStack etc? For example:

```swift
VStack(alignment: .trailing) {
  Section("Header") {
    Text("Content")
  }
}
```

Section on iOS behaves more like Group so "Header" and "Content" will be trailing aligned.

aabewhite: What does SwiftUI do if you put a .border or some other modifier on the Section?

I'm OK with having undefined layout outside of a container - Skip doesn't generally support SwiftUI's implicit VStack behavior. But I'd like modifiers to behave similarly.

ky-is: This doesn't change the current behavior of not directly applying modifiers to the Section (Section is a Group not a container on iOS). `.border` on a iOS Section applies individually so each child (including the header/footer) will get a separate border. Environment modifiers like foregroundStyle work as expected.

PR #124, opened 2025-02-11
Add labelStyle support

ky-is:
Skip Pull Request Checklist:

### DIFF
--- a/Sources/SkipUI/SkipUI/Text/Label.swift
+++ b/Sources/SkipUI/SkipUI/Text/Label.swift
@@ -46,35 +46,41 @@ public struct Label : View {
 
     #if SKIP
     @Composable public override func ComposeContent(context: ComposeContext) {
+        var style = EnvironmentValues.shared._labelStyle ?? LabelStyle.automatic
         let placement = EnvironmentValues.shared._placement
-        if placement.contains(ViewPlacement.toolbar) {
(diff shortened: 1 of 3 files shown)


Sources/SkipUI/SkipUI/Text/Label.swift
```diff
@@ -46,35 +46,41 @@ public struct Label : View {
...
-            if let imageColor {
-                EnvironmentValues.shared.setValues {
-                    $0.set_foregroundStyle(imageColor)
-                } in: {
-                    image.Compose(context: context.content())
-                }
-            } else {
-                image.Compose(context: context.content())
-            }
+            ComposeImage(context: context, imageColor: imageColor)
```
aabewhite, line 83: I'd like to do context.content() here and use the given Context as-is in ComposeImage(), because ComposeImage() *should* apply the full context if called directly from the style switch or is called by some other component (it's not private)

aabewhite (changes requested):
One comment, otherwise looks great!

Sources/SkipUI/SkipUI/Text/Label.swift
aabewhite: Oh, one other thing: it'd be great if you added .labelStyle support to the supported API table in the README

aabewhite (approved):

PR #125, opened 2025-02-12
Fix Menu Link Label formatting

### LINKED ISSUE #121: Bug: ShareLink not showing text for Label in a .toolbar Menu
### Description
The following code will show the icon but not the title:
```
List {
    ...
}.toolbar {
    ToolbarItem {
        Menu {
            Button { } label: {
                Label("Settings", systemImage: "gearshape")
            }
            ShareLink(item: URL("https://apple.com")!) {
                Label("Share", systemImage: "square.and.arrow.up")
            }
        } label: {
            Image(systenName: "ellipsis")
        }
    }
}
```

ky-is:
Because Links wrap a Button, they are not currently recognized by Menu and the custom Button layout is not applied. Also, Menus inside a toolbar apply the icon only style which is incorrect inside a Menu. This fixes #121.

This implementation is kind of ugly as it relies on calling ComposeAction to prepare the Intent/openURL handler. Ideas on a better approach?

Slightly unrelated but I also added missing Button initializers with systemImage that create a Label.

---


aabewhite (approved):
LGTM!

PR #137, opened 2025-03-02
Fix deprecated ProgressView for Android

pacient:
There seems to be a deprecated use of Android's `LinearProgressIndicator`. I did a fix to look same as it did before for when there is a value. Let me know if I should update to include same change for when the progress is indeterminate.

Before:
Screenshot 2025-03-02 at 11 32 44

After:
Screenshot 2025-03-02 at 11 40 28

------

Skip Pull Request Checklist:

- [x] REQUIRED: I have signed the [Contributor Agreement](https://github.com/skiptools/clabot-config) https://github.com/skiptools/clabot-config/pull/31 

### DIFF
--- a/Sources/SkipUI/SkipUI/Components/ProgressView.swift
+++ b/Sources/SkipUI/SkipUI/Components/ProgressView.swift
@@ -95,7 +95,13 @@ public struct ProgressView : View {
         if value == nil || total == nil {
             LinearProgressIndicator(modifier: modifier, color: color)
         } else {
-            LinearProgressIndicator(progress: Float(value! / total!), modifier: modifier, color: color)
+            LinearProgressIndicator(
+                progress: { Float(value! / total!) },
+                modifier: modifier,
+                color: color,
+                gapSize = -8.dp, // Makes sure the progress overlaps the background
+                drawStopIndicator = {}
+            )
         }
     }
 


marcprux:
Why -8.dp? This does change the appearance of the progress indicator, like the bottommost one in your screenshot where the "Before" indicator has a visible gap, and the "After" one doesn't.

Also, if you are just trying to get rid of a deprecation warning, could we just use this constructor instead: https://developer.android.com/reference/com/google/android/material/progressindicator/LinearProgressIndicator#LinearProgressIndicator(android.content.Context,%20android.util.AttributeSet)

pacient:
Looking at the [ProgresView](https://skip.tools/docs/components/progressview/) page I could see that the way progress indicator looked like is different. I added -8.dp so it over laps the progress with the background, otherwise you still see a bit of a gap even with 0, maybe It could be reduced to 4. Also, it's not about the deprecation warning. It's also the dot at the end, which `drawStopIndicator` fixes. With the changes the ProgressView matches how the iOS looks like

Sources/SkipUI/SkipUI/Components/ProgressView.swift
```diff
@@ -95,7 +95,13 @@ public struct ProgressView : View {
         if value == nil || total == nil {
             LinearProgressIndicator(modifier: modifier, color: color)
         } else {
-            LinearProgressIndicator(progress: Float(value! / total!), modifier: modifier, color: color)
+            LinearProgressIndicator(
+                progress: { Float(value! / total!) },
```
aabewhite, line 99: Thanks for the deprecated API fix. I'd like to leave everything else as using default values (i.e. just exclude the gapSize and drawStopIndicator parameters) for the normal Compose look. We can eventually add custom modifiers to change the Material view, as we've done for other components.

And could you also make the fix for CircularProgressIndicator?

aabewhite (changes requested):

PR #147, opened 2025-03-10
Allow tagging with optional types

ky-is:
SwiftUI lets you `.tag(nil as String?)`. This is useful for providing a default value for a Picker where the selection value is a `String?` in this case.

I'm not sure if more should be done in ForEach, where `value` is guarded non-nil before being passed to TagModifierView: https://github.com/skiptools/skip-ui/blob/63bc24afa860656c34dba4b90146680c8d3aa74b/Sources/SkipUI/SkipUI/Containers/ForEach.swift#L227

---


aabewhite (approved):
Thank you!

PR #152, opened 2025-03-17
Autocomplete disabled when isSecure == true

m-t-a-p:
A fix for the SecureField on Android - autocomplete feature disabled to secure the password.
Related discussion here: https://skiptools.slack.com/archives/C078X69G8F2/p1741698205560339


marcprux:
Thanks!

PR #161, opened 2025-03-24
Add support for lineLimit(_ limit: Int, reservesSpace: Bool)

dfabulich:
https://developer.apple.com/documentation/swiftui/view/linelimit(_:reservesspace:)

I'm filing this as a draft PR because I _think_ it should be working, but it isn't working in fact.


dfabulich:
I think a Tuple might be nicer, but I'm not sure it would make sense here, because `.lineLimit` is a built-in official environment value of SwiftUI, and it's an `Int`.

https://developer.apple.com/documentation/swiftui/environmentvalues/linelimit 

The current released version of SkipUI can honor the `.lineLimit` environment variable today; it'll just work. If we changed the type of `.lineLimit`, it would break anyone who sets `.environment(\.lineLimit, 2)`.

And if we _don't_ change the type, then we just have to have a second environment variable. (Right?)

aabewhite:
Ah, I didn't notice that lineLimit is one of the exposed environment values. In that case you're absolutely right to keep them separate

Sources/SkipUI/SkipUI/Environment/EnvironmentValues.swift
```diff
@@ -282,6 +282,11 @@ extension EnvironmentValues {
         set { setBuiltinValue(key: "lineLimit", value: newValue, defaultValue: { nil }) }
     }
 
+    public var lineLimitReservesSpace: Bool? {
```
aabewhite, line 285: This is not a public EnvironmentValues key in SwiftUI (correct me if I'm wrong), so it should go below in the "Internal values" section and be underscore-prefixed

aabewhite (commented):
One request, otherwise looks good!

PR #189, opened 2025-04-26
Consider closures equal when comparing preference values to avoid recoups

aabewhite:
https://github.com/skiptools/skip-ui/issues/188


marcprux:
When is a preference a closure?

PR #196, opened 2025-05-03
Include example images in the README

dfabulich:
There are a number of subtle variations between the fallback symbols for SF Symbols and the Material Icons. In this PR, I've checked in the actual images from SF Symbols and Material Icons, so they can be displayed in the README.


marcprux:
We don't want to bring in all those symbols to this repository, both because it inflates the size of the checkout that every Skip user will experience when building skip-ui, and also because there's not much point in listing all the symbols only to say that the very few, rough and minimal fallbacks that we provide do not match the iOS images.

We will address this better in the documentation, following up in https://github.com/skiptools/skip-ui/issues/200

PR #210, opened 2025-05-21
onPreferenceChange

### LINKED ISSUE #133: Implement .onPreferenceChange...
I have a fairly complex layout scenario which would benefit from the following technique (scroll down to "Using a preference key"):

https://swiftbysundell.com/questions/syncing-the-width-or-height-of-two-swiftui-views/

I use this in my iOS code and it works well.

Unfortunately, while Skip does seem to support `PreferenceKey` generally, `.onPreferenceChange` is not supported. This prevents me from being able to get the CGSize of a view and use it for calculations as needed.

Would it be possible to implement `.onPreferenceChange`?

dfabulich:
Fixes #133

This PR is weird in several ways.

First, it appears that `SKIP DECLARE` won't let me disregard the complex `where` clause of the declaration of this function. If I uncomment the `K.Value : Equatable` part, Skip will complain, "Skip does not support the referenced type as a generic constraint." I would have expected `SKIP DECLARE` to tell Skip to disregard that problem, but it doesn't. (Relaxing that constraint should be harmless in practice, because the Xcode build will reject it if someone tries to make a PreferenceKey of a non-Equatable value.) I've filed this as a bug on the transpiler. https://github.com/skiptools/skip/issues/432

Second, I'm using `snapshotFlow`; I have no idea how I'd port that to Swift, so I just didn't; I'm heavily using `SKIP INSERT` for this.

Third, you can't just build out an ordinary `PreferenceKey` and use it with `preference`. I had to declare my PreferenceKey like this:

```swift
struct TestPreferenceKey: PreferenceKey {
    typealias Value = String
    
    #if SKIP
    // SKIP DECLARE: companion object: PreferenceKeyCompanion<String>
    final class Companion: PreferenceKeyCompanion {
        let defaultValue = ""
        func reduce(value: inout String, nextValue: () -> String) {
            value = nextValue()
        }
    }
    #else
    static var defaultValue: String = ""

    static func reduce(value: inout String, nextValue: () -> String) {
        value = nextValue()
    }
    #endif
}
```

That is, in SwiftUI, I could just declare a couple of static methods, but in SkipUI, I had to explicitly declare a companion object so I could declare it a `PreferenceKeyCompanion`.

This is all quite strange, but it _does_ work. (And isn't that really what Skip is all about? 😉)

Skip Pull Request Checklist:

### DIFF
--- a/Sources/SkipUI/SkipUI/Environment/PreferenceKey.swift
+++ b/Sources/SkipUI/SkipUI/Environment/PreferenceKey.swift
@@ -5,17 +5,22 @@
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.ProvidedValue
 import androidx.compose.runtime.SideEffect
 import androidx.compose.runtime.Stable
 import androidx.compose.runtime.compositionLocalOf
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.runtime.setValue
+import androidx.compose.runtime.snapshotFlow
 import androidx.compose.runtime.saveable.Saver
 import androidx.compose.runtime.saveable.rememberSaveable
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.distinctUntilChanged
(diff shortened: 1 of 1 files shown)


marcprux:
> Any feedback on this?

I'm hoping to get @aabewhite's feedback but he's on vacation right now.

dfabulich:
I see that this branch is now conflicted, but I'd appreciate some feedback from @aabewhite before I bother trying to rebase it. Will you accept a PR like this at all? (This PR has been on hold for quite a while.)

aabewhite:
Thanks for this. I plan to use this as a base for supporting custom preference keys and onPreferenceChanged soon. No need to rebase it

aabewhite:
Incorporated into https://github.com/skiptools/skip-ui/pull/242

PR #216, opened 2025-05-27
Add support for accessibilityHint(_:)

dfabulich:
I guess this isn't a perfect match for <https://developer.apple.com/documentation/swiftui/view/accessibilityhint(_:)>, because I think the iOS version applies whenever the user "performs the view's action," whereas my implementation only applies when the user clicks on a clickable (a button).

But I've never seen anyone use accessibility hints on anything except buttons. (I'm struggling to imagine how/why an accessibility hint would be necessary on anything _but_ a button.)

If you think this implementation isn't close enough to iOS's behavior, we could document that users should do this instead, at least for buttons:

```swift
#if SKIP
import androidx.compose.ui.semantics.onClick
import androidx.compose.ui.semantics.semantics
#endif

Button("Click me") { print("clicked") }
#if SKIP
.composeModifier { $0.semantics { onClick(label: "Insert hint here", action: nil) } }
#else
.accessibilityHint("Insert hint here")
#endif
```

But IMO that's even jankier.

Skip Pull Request Checklist:

### DIFF
--- a/Sources/SkipUI/SkipUI/System/Accessibility.swift
+++ b/Sources/SkipUI/SkipUI/System/Accessibility.swift
@@ -6,6 +6,7 @@ import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.semantics.contentDescription
 import androidx.compose.ui.semantics.heading
 import androidx.compose.ui.semantics.invisibleToUser
+import androidx.compose.ui.semantics.onClick
 import androidx.compose.ui.semantics.popup
 import androidx.compose.ui.semantics.role
 import androidx.compose.ui.semantics.selected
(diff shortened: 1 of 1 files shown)


marcprux:
I'm confused by the meaning of [SemanticsPropertyReceiver.onClick](https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/package-summary#(androidx.compose.ui.semantics.SemanticsPropertyReceiver).onClick(kotlin.String,kotlin.Function0)). What does the action parameter actually do? Is it ever invoked, and if so, will passing `null` cause the button's behavior to change?

dfabulich:
The `action` parameter is invoked only by accessibility services, e.g. TalkBack. It's a `() -> Boolean`; if it returns `true`, then the semantics `onClick` handler stops event propagation, preventing the button from doing its normal `onClick` action. If it returns `false` (or if you pass a null action), then the event propagates to the button, and its normal `onClick` handler fires.

```
Button(
    modifier = Modifier
        .semantics {
            onClick(label = "Click", action = {
                Log.i("onClick", "semantic click")
                return false
            })
        },
    onClick = {
        Log.i("onClick", "regular click")
    }) {
    Text("Button")
}
```

* If you click that button normally, it will just log "regular click."
* If you turn on TalkBack and activate the button by double tapping, it will log "semantic click" (twice, for some reason…?) and then log "regular click".
* If you `return true` instead of `return false`, it will _only_ log "semantic click".
* If you set `action = null` and activate via TalkBack, it will log "regular click".

In this PR, we just want to set the label, so setting `action = null` will do the right thing.

PR #235, opened 2025-07-13
Enable toolbarTitleMenu in skip-ui

woodymelling:
Enable the .toolbarTileMenu in SkipUI. This is the minimum to get something functional for my project, I'm not sure what else would be required to get it over the line. I tried using the already exist `class Menu: View` but I couldn't get it working. Instead we'll just grab the tooling that makes the menuItems. This required the ToolbarContent to change to use a ComposeBuilder. There might be a better way to approach this so the ToolbarContent doesn't change.

Skip Pull Request Checklist:


aabewhite:
This functionality has now been merged in a separate PR that I based on the code here. Thanks!

PR #236, opened 2025-07-21
Refactor to fix edge cases around modifiers and custom views

aabewhite:
In general, changes pipeline from:
View -> Compose() -> ComposeContent()

to:
View -> Evaluate() -> Renderables -> Render()

This accomplishes two things:
1. Allows users to wrap content in custom views in places they couldn't before. For example, a List can now contain custom views that define Sections, where previously any Section would have to be inlined directly into the List content.
2. Allows for modifiers on views that evaluate to multiple elements, such as ForEach.

If we put in some work, we can also add support for e.g. SwiftUI's implicit VStack behavior when you define multiple views in the body of a custom view.


aabewhite:
DO NOT MERGE. This is in progress

PR #253, opened 2025-09-08
Use ImageDecoder for UIImage loading from data and files to handle correct rotation

GavynHolt:

Skip Pull Request Checklist:

### DIFF
--- a/Sources/SkipUI/SkipUI/UIKit/UIImage.swift
+++ b/Sources/SkipUI/SkipUI/UIKit/UIImage.swift
@@ -8,7 +8,6 @@ import android.graphics.ImageDecoder
 import android.net.Uri
 import java.nio.ByteBuffer
 import android.graphics.Bitmap
-import java.io.ByteArrayInputStream
 #endif
 
 // SKIP @bridge


Sources/SkipUI/Skip/skip.yml
```diff
@@ -82,6 +82,7 @@ build:
         - 'api(libs.androidx.appcompat.resources)'
         - 'api(libs.androidx.activity.compose)'
         - 'api(libs.androidx.lifecycle.process)'
+        - 'implementation("androidx.exifinterface:exifinterface:1.4.1")'
```
marcprux, line 85: It might be better to instead add a dependency on [coil-core-android](https://mvnrepository.com/artifact/io.coil-kt.coil3/coil-core-android) (which appears to depend on [exifinterface](https://mvnrepository.com/artifact/androidx.exifinterface/exifinterface), and so will bring it in transitively). That way, you can just use `versionRef("coil")` to make keep everything up to date with the rest of the coil dependencies.

marcprux:
This looks good, but I'm wondering if you can try one more thing: switching from `BitmapFactory.decodeByteArray` / `BitmapFactory.decodeFile` to the newer[^1] [`ImageDecoder.decodeBitmap`](https://developer.android.com/reference/kotlin/android/graphics/ImageDecoder#decodeBitmap(android.graphics.ImageDecoder.Source)). As well as supporting more formats and being (allegedly) higher-performance, I've read somewhere that it automatically handles correct rotation. Since the rotation would be done as part of the image loading, it would be higher performance than adding it as a post-load step, which would help reduce the chances of ANRs when loading large images.

Even if it doesn't handle the rotation, it would probably be good to switch to using it anyway and then keeping in your post-load rotation-from-exif-metadata logic.

[^1]: New as of API 28, which coincidentally is Skip's minimum supported AP level

marcprux:
Glad that worked!

PR #281, opened 2025-12-21
Bugfix: Status bar does not update when a sheet is presented on Android

fhasse95:
Currently, when a sheet is presented on Android, the status bar style does not update accordingly. Instead of using black icons / text in light mode and white icons / text in dark mode, the icons always use the light mode variant, making the status bar basically invisible in dark mode.

Steps to Reproduce:
- Run the "Skip Showcase" app on Android
- Navigate to "Showcase" => "Sheet" and open any of the examples
- Switch the system appearance from light to dark mode (or vice versa)
- Observe the status bar icons
  => The status bar icons always use the light mode variant (dark icons / text) and do not adapt to the system theme ⚡

This PR solves the issue by updating the status bar style of the dialog window that presents the sheet depending on the current system appearance (similar to this [PR](https://github.com/skiptools/skipapp-showcase/pull/50)).

**Example (Before):**
Before

**Example (After):**
After

---

Skip Pull Request Checklist:

### DIFF
--- a/Sources/SkipUI/SkipUI/Layout/Presentation.swift
+++ b/Sources/SkipUI/SkipUI/Layout/Presentation.swift
@@ -38,7 +38,7 @@ import androidx.compose.material3.Surface
 import androidx.compose.material3.rememberModalBottomSheetState
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
-import androidx.compose.runtime.key
+import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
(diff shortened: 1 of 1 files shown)


fhasse95:
Hi @marcprux,
any updates on this?

I’m happy to help or make additional adjustments if needed to get this merged :)

marcprux:
Apologies for not following-up on this. This method (similar to the improvement you contributed at https://github.com/skiptools/skipapp-showcase/pull/50) does work, but I wonder if it might be better to use the [ModalBottomSheetProperties](https://developer.android.com/reference/kotlin/androidx/compose/material3/ModalBottomSheetProperties#ModalBottomSheetProperties(kotlin.Boolean,kotlin.Boolean,androidx.compose.ui.window.SecureFlagPolicy,kotlin.Boolean,kotlin.Boolean))'s `isAppearanceLightStatusBars` and `isAppearanceLightNavigationBars` properties (newly introduced in [compose-material3 1.4.0](https://developer.android.com/jetpack/androidx/releases/compose-material3#1.4.0-alpha13)), which seems custom-tailored for this particular scenario.

Let us know what you think. If `ModalBottomSheetProperties` doesn't work for this, then we can go ahead with this change.

marcprux:
WDYT @aabewhite?

PR #298, opened 2026-01-12
Add Android modifier for logging layout

dfabulich:
I use this thing _constantly_. I keep adding it back in and ripping it out of SkipUI every time I want to work on a tricky layout issue; I figure I might as well file a PR for it. You use it like this: `modifier.logLayout("MyLayoutTag")` and it will log the incoming layout constraints from the parent component, as well as the actual bounds when the component is positioned.

It's way better than just slapping on a red border!

Skip Pull Request Checklist:

### DIFF
--- a/Sources/SkipUI/SkipUI/View/AdditionalViewModifiers.swift
+++ b/Sources/SkipUI/SkipUI/View/AdditionalViewModifiers.swift
@@ -26,6 +26,11 @@ import androidx.compose.ui.graphics.BlendMode
 import androidx.compose.ui.graphics.CompositingStrategy
 import androidx.compose.ui.graphics.TransformOrigin
 import androidx.compose.ui.graphics.graphicsLayer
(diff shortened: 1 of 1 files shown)


Sources/SkipUI/SkipUI/View/AdditionalViewModifiers.swift
```diff
@@ -1201,6 +1206,32 @@ extension View {
 }
 
 #if SKIP
+extension Modifier {
+    /// Log layout constraints for debugging purposes.
+    ///
+    /// - Parameter tag: The log tag to use (default: "LogLayout").
+    /// - Returns: A modifier that logs layout constraints.
+    func logLayout(tag: String = "LogLayout") -> Modifier {
```
marcprux: That looks handy!

Could you add a blurb in the README about how to use it, and maybe some sample output?

marcprux (changes requested):

PR #330, opened 2026-02-09
Fix .id to reset state when it changes

aabewhite:
Fix for https://github.com/skiptools/skip-ui/issues/274


aabewhite:
This is a little dangerous because it affects the state tracking of anyone using .id modifier, and it isn't super well tested

marcprux:
This seems to work well in my experimentation!

PR #358, opened 2026-03-17
Fix default scaling of images in tab view tabs

marcprux:
Previously, symbol/vector images would be rendered with a default scaling of 1.5 in order to accommodate a (common) case of using symbol images in tab views. This would get the image sized about right, but would make it look "heavier" than its iOS equivalent, as can be seen in the previous Showcase center tabs (which is the only one that uses a custom symbol image):

Screenshot_1773719375 copy
Simulator Screenshot - iPhone 17 - 2026-03-17 at 18 22 59

With this changes, the image itself will not be scaled by by default, but when an image is being rendered in a tab it will apply the font scaling of 1.5, as well as defaulting to a light font weight in order to replicate the default iOS tab bar behavior.

The resulting tab icons are much closer:

Screenshot_1773786183
Simulator Screenshot - iPhone 17 - 2026-03-17 at 18 22 59

### DIFF
--- a/Sources/SkipUI/SkipUI/Components/Image.swift
+++ b/Sources/SkipUI/SkipUI/Components/Image.swift
@@ -395,8 +395,7 @@ public struct Image : View, Renderable, Equatable {
                 let textSizeDp = with(LocalDensity.current) {
                     textStyle.fontSize.toDp()
                 }
-                // Apply a multiplier to more closely match SwiftUI's relative text and system image sizes
-                modifier = Modifier.size(textSizeDp * Float(1.5))
+                modifier = Modifier.size(textSizeDp)
             } else {
                 modifier = Modifier
             }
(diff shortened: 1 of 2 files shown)


djch:
@marcprux I noticed in your screenshots (and my app) the tabs lost some padding inside the 'oval' shape around the selected tab's icon. Was that intentional?

djch:
From my perspective, it looked better with the padding. I think Jetpack Compose's documentation has them with padding too:
navigation-bar
https://developer.android.com/develop/ui/compose/components/navigation-bar

marcprux:
> From my perspective, it looked better with the padding.

I am inclined to agree. Would you mind filing a [new issue](https://github.com/skiptools/skip-ui/issues/new) about this so we don't lose track of it?

And, of course, contributions are always welcome 😉

dfabulich:
I've filed it as #367

djch:
Thanks for doing that @dfabulich 🙇‍♂️

PR #372, opened 2026-03-30
Test tags

dfabulich:
These test tags help UI Automator to find the tab bar and the search field.

Skip Pull Request Checklist:

-----

-----

### DIFF
--- a/Sources/SkipUI/SkipUI/Commands/Search.swift
+++ b/Sources/SkipUI/SkipUI/Commands/Search.swift
@@ -29,7 +29,10 @@ import androidx.compose.ui.input.nestedscroll.NestedScrollConnection
 import androidx.compose.ui.input.nestedscroll.NestedScrollSource
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.platform.LocalFocusManager
+import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.semantics.testTagsAsResourceId
 import androidx.compose.ui.text.input.ImeAction
 import androidx.compose.ui.unit.dp
 import kotlinx.coroutines.launch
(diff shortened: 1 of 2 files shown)


dfabulich:
This will conflict with #387. (The failing tests are just flaky in this case.)

marcprux:
Seems useful. 

For the future, is there any reason we would _not_ want `semantics { testTagsAsResourceId = true }`? We also set this from `accessibilityIdentifier`, so I wonder if we should just have this modifier be the default for all SkipUI views…

https://github.com/skiptools/skip-ui/blob/b027e538a68321b90cff19b62936993acf65ec8c/Sources/SkipUI/SkipUI/System/Accessibility.swift#L24-L32

dfabulich:
I'm a little hazy on this myself.

tl;dr: We probably always want `semantics { testTagsAsResourceId = true }`, but maybe we should feel a little bad about that, and maybe stop doing it that way some day.

We probably always want it because [UI Automator](https://developer.android.com/training/testing/other-components/ui-automator) can only use resource IDs. https://github.com/dfabulich/skip-ui-automation/ is built on UI Automator. [Compose Testing](https://developer.android.com/develop/ui/compose/testing) can access the underlying test tags. 

So, why would we feel bad about it? 

Partly because test tags can have a many-to-many relationship with elements on the Compose semantic tree. You can assign the same test tag to multiple elements, and any given element can have several different test tags.

By contrast, individual elements are allowed at most one resource ID, and they're kinda intended to be Unique™. I say "kinda intended" because nothing _crashes_ if you reuse the same resource ID on one Android screen, and UI Automator _can_ handle it if you have two elements with the same ID, but the best practice is for each ID to be used once per screen.

Resource IDs are supposedly a "legacy" thing, perhaps because UI Automator is, itself, aspirationally a legacy thing. (Resource IDs are based on the old IDs used in Android XML views.) But, based on my research, there's lots of stuff that you can _only_ do with UI Automator that you can't do with Compose Testing.

Maybe one day UI Automator will be able to access semantics tags; at that point, we could probably stand to use the test tags directly without trying to convert them to resource IDs. But that might be very far in the future!

Issue #370, opened 2026-03-28
Add support for predictive back


dfabulich:
https://developer.android.com/develop/ui/compose/system/predictive-back
https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture
https://developer.android.com/design/ui/mobile/guides/patterns/predictive-back

We're currently using a `BackHandler`, which prevents predictive back animations.

https://github.com/skiptools/skip-ui/blob/b98d3a21fd25714db6243dbbde7ea1b1240e522a/Sources/SkipUI/SkipUI/Containers/Navigation.swift#L261-L266

marcprux:
Fixed by https://github.com/skiptools/skip-ui/pull/377

PR #377, opened 2026-03-31
Navigation 3, predictive back

### LINKED ISSUE #292: Investigate migrating to Compose Navigation 3
See: 
https://developer.android.com/guide/navigation/navigation-3/migration-guide
https://www.youtube.com/watch?v=6L6GfvlmKj8

dfabulich:
Fixes #292 and #370

Relevant Google docs:

* Navigation 2 (I had to brush up on Nav2 to do this PR. Maybe you need a refresher?)
  * https://developer.android.com/guide/navigation
  * https://developer.android.com/develop/ui/compose/navigation
* https://developer.android.com/guide/navigation/navigation-3
  * ^^ the whole thing, but especially https://developer.android.com/guide/navigation/navigation-3/migration-guide
* Predictive back
  * https://developer.android.com/develop/ui/compose/system/predictive-back
  * https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture
  * https://developer.android.com/design/ui/mobile/guides/patterns/predictive-back

My goal was to keep this PR as lightweight as possible, to make it as straightforward as possible to review it. (I recommend viewing it in a whitespace-ignoring diff tool, even just `git diff -w`.)

I think a deeper refactoring is possible here, to integrate more deeply with Navigation 3, but I figure we'd land this, shake out bugs, and then refactor in a subsequent PR (if desired).

# `Navigation.swift`

The migration guide calls for these steps:
> 1. Add the Navigation 3 dependencies.
> 2. Update your navigation routes to implement the `NavKey` interface.
> 3. Create classes to hold and modify your navigation state.
> 4. Replace `NavController` with these classes.
> 5. Move your destinations from `NavHost`'s `NavGraph` into an `entryProvider`.
> 6. Replace `NavHost` with `NavDisplay`.
> 7. Remove Navigation 2 dependencies.

I'm going to explain these steps slightly out of order.

* I have nothing interesting to say about Step 1 or Step 7; I updated `skip.yml`.
* Steps 2, 3, and 5 are about the new Navigation 3 data model, which I'll summarize below.
* With the new data model in place, all that remained were steps 4 and 6, which were mindless replacements.

## `NavKey`s instead of routes (Step 2, Step 3, Step 5)

In Nav2 terms, `Navigation.swift`'s `Navigator` routes were strings, either `Navigator.rootRoute` `"navigationroot"` or a route defined by the static `route(for:valueString:)` function.

```swift
/// Route for the given destination index and value string.
static func route(for destinationIndex: Int, valueString: String) -> String {
    return String(describing: destinationIndex) + "/" + valueString
}
```

In the `NavigationStack` `@Composable Render(context:)`, we defined the navigation graph in code, passing the root composable, followed by a list of composables (each with its own destination route) to the `NavHost` constructor.

In this PR, I defined two `NavKey` objects:

```swift
// SKIP INSERT: @Serializable
public enum SkipNavigationStackRootKey : NavKey {
    case root
}

// SKIP INSERT: @Serializable
public struct SkipNavigationStackPushKey : NavKey {
    public let destinationIndex: Int
    public let identifier: String
}
```

`@Composable Render(context:)` now builds an "entry provider" using these two types; no need for a for loop over destinations here. We still use `route(for:valueString:)` to construct a stable entry ID for each entry in the navigation stack:

```swift
private func stableEntryId(forKey key: NavKey) -> String {
    if key is SkipNavigationStackRootKey {
        return Self.rootRoute
    } else if let pushKey = key as? SkipNavigationStackPushKey {
        return Self.route(for: pushKey.destinationIndex, valueString: pushKey.identifier)
    } else {
        return String(describing: key)
    }
}
```

`Navigator` kept a `private var backStackState: [String: BackStackState] = [:]` … we're still using it, with the `stableEntryId` as the key.

```swift
func state(forPushKey pushKey: SkipNavigationStackPushKey) -> BackStackState? {
    let id = stableEntryId(forKey: pushKey)
    return backStackState[id]
}
```

## Replacing `NavController` with `NavBackStack<NavKey>` and replacing `NavHost` with `NavDisplay` (Steps 4 and 6)

Instead of a `NavController`, Navigation 3 uses a "back stack", which can be a simple array, but they recommend using `rememberNavBackStack` which creates a `NavBackStack<NavKey>`. Everything you would do with a `navController`, you now do by mutating the `navBackStack`, so there's a lot of find-and-replace of `navController` with `navBackStack` throughout.

In Navigation 2, we included a `BackHandler` which called `navigator.value.navigateBack()`; this broke predictive back. In Navigation 3, `NavDisplay` replaces `NavHost`, and includes an `onBack` callback where we can call `navigator.value.navigateBack()`. We therefore no longer require a `BackHandler`. By simply deleting it, we get predictive-back animations for free.

In fact, we get _all_ stack transition animations for free, so I just deleted the existing animation definitions, too, preferring to use Android's recommended defaults. (Perhaps in a subsequent PR, we'll implement the [`navigationTransition(_:)`](https://developer.apple.com/documentation/swiftui/view/navigationtransition%28_%3A%29) modifier, allowing users to customize the transition animation for individual stack entries. Luckily for me, we don't support that yet, so I don't have to worry about it right now.)

# `TabView.swift`

The `TabView` `NavHost` did a thing that strikes me as quite strange: it defined a navigation graph of 100 tabs, identified by ID number. There's a comment explaining: "Use a constant number of routes. Changing routes causes a NavHost to reset its state"

I didn't want to mess with this, and I wasn't at all sure how to do regression testing against ripping that out, so I made a little function to return a list of 100 `NavBackStack`s, leaving in a little "TODO" comment:

```swift
/// Fixed 100 persistent back stacks for tab bar content (matches legacy `0..<100` routes).
@Composable public func rememberSkipTabViewBackStacks() -> kotlin.collections.List<NavBackStack<NavKey>> {
    // Use a constant number of routes. Changing routes causes a NavHost to reset its state
    // TODO is this necessary with NavDisplay?
    let slots = remember { arrayOfNulls<NavBackStack<NavKey>?>(100) }
    var tabIndex = 0
    while tabIndex < 100 {
        let ti = tabIndex
        key(ti) {
            slots[ti] = rememberNavBackStack(SkipTabViewRouteKey(index: ti))
        }
        tabIndex += 1
    }
    return slots.filterNotNull()
}
``` 

And we also created a new `NavKey` route for `TabView`, tracking the tab index.

```swift
// SKIP INSERT: @Serializable
public struct SkipTabViewRouteKey : NavKey {
    public let index: Int
}
```

`TabView`s `NavDisplay` then has a generic handler for any `SkipTabViewRouteKey`.

`TabView` now tracks its own `selectedTabIndex` state instead of asking `navController` to navigate to a numbered route. Functions that accepted a `navController` now accept `selectedTabIndex: MutableState<Int>` and the `tabBackStacks`.

What `TabView` called a `route` is now just a stringified `tabIndex`. (It probably doesn't need to be stringified, but it kept this huge PR from being even huger, because we already had a function that converts string routes into tab tag values.)

Skip Pull Request Checklist:

-----

The migration guide recommended prompting an AI to do the transition, "Migrate from Navigation 2 to Navigation 3 using the official migration guide." I started with that, and cleaned up and thoroughly reviewed the result.

To verify the changes, I wrote automated tests! ✨ https://github.com/skiptools/skipapp-showcase/pull/83

…and I also just clicked around the NavigationStackPlayground and the TabView playground in the showcase.

## Manually testing predictive back

I especially enjoyed testing predictive back. It's a little thorny to set it up in the emulator. I created an API 36 (Android 16) medium phone with the Google Play Store. I searched settings for "Gesture Navigation" and turned on "Gesture navigation." With that on, if you drag on the emulator from the left 5% of the screen or the right 5% of the screen toward the center, you'll see a predictive back animation.

Prior to this PR, the Showcase already supported the built in "back-to-home" predictive back animation, but when navigating back from a playground to the list, nothing animated until you released your finger. But now…

[Screen_recording_20260331_154710.webm](https://github.com/user-attachments/assets/3ae27fdb-7876-41a1-b032-3e5cf6d4dcdb)

### DIFF
--- a/Sources/SkipUI/Skip/skip.yml
+++ b/Sources/SkipUI/Skip/skip.yml
@@ -13,7 +13,8 @@ settings:
             - block: 'create("libs")'
               contents:
                 - 'version("coil", "3.4.0")'
-                - 'version("androidx-navigation", "2.9.7")'
+                - 'version("androidx-nav3", "1.0.1")'
+                - 'version("kotlinx-serialization-json", "1.8.1")'
                 - 'version("androidx-appcompat", "1.7.1")'
                 - 'version("androidx-activity", "1.13.0")'
                 - 'version("androidx-lifecycle-process", "2.10.0")'
@@ -32,7 +33,10 @@ settings:
                 - 'library("androidx-appcompat", "androidx.appcompat", "appcompat").versionRef("androidx-appcompat")'
                 - 'library("androidx-appcompat-resources", "androidx.appcompat", "appcompat-resources").versionRef("androidx-appcompat")'
 
-                - 'library("androidx-navigation-compose", "androidx.navigation", "navigation-compose").versionRef("androidx-navigation")'
+                - 'library("androidx-navigation3-runtime", "androidx.navigation3", "navigation3-runtime").versionRef("androidx-nav3")'
+                - 'library("androidx-navigation3-ui", "androidx.navigation3", "navigation3-ui").versionRef("androidx-nav3")'
+                - 'library("kotlinx-serialization-json", "org.jetbrains.kotlinx", "kotlinx-serialization-json").versionRef("kotlinx-serialization-json")'
+                - 'plugin("kotlin-serialization", "org.jetbrains.kotlin.plugin.serialization").versionRef("kotlin")'
                 - 'library("androidx-activity-compose", "androidx.activity", "activity-compose").versionRef("androidx-activity")'
                 - 'library("androidx-lifecycle-process", "androidx.lifecycle", "lifecycle-process").versionRef("androidx-lifecycle-process")'
                 - 'library("androidx-compose-material3-adaptive", "androidx.compose.material3.adaptive", "adaptive").versionRef("androidx-material3-adaptive")'
@@ -54,6 +58,7 @@ build:
     - block: 'plugins'
       contents:
         - 'alias(libs.plugins.kotlin.compose)'
+        - 'alias(libs.plugins.kotlin.serialization)'
 
     - block: 'android'
       contents:
@@ -73,7 +78,9 @@ build:
         - 'api(libs.androidx.compose.material3)'
         - 'api(libs.androidx.compose.material3.adaptive)'
         - 'api(libs.androidx.compose.foundation)'
-        - 'api(libs.androidx.navigation.compose)'
+        - 'api(libs.androidx.navigation3.runtime)'
+        - 'api(libs.androidx.navigation3.ui)'
+        - 'api(libs.kotlinx.serialization.json)'
         - 'api(libs.androidx.appcompat)'
         - 'api(libs.androidx.appcompat.resources)'
         - 'api(libs.androidx.activity.compose)'
--- a/Sources/SkipUI/SkipUI/Containers/Navigation.swift
+++ b/Sources/SkipUI/SkipUI/Containers/Navigation.swift
@@ -8,6 +8,7 @@ import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
 import androidx.compose.animation.slideInHorizontally
 import androidx.compose.animation.slideOutHorizontally
+import androidx.compose.animation.togetherWith
 import androidx.compose.animation.core.tween
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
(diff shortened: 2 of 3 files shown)


marcprux (approved):
This seems solid in all my testing. Overall, a great (and necessary) improvement, thanks!

PR #398, opened 2026-04-17
TabView: Keep a stable SaveableStateHolder for each tab's backStack

### LINKED ISSUE #397: TabView forgets scroll offset when switching tabs
In the Showcase app, tap on the Showcase tab and scroll down a bit. Switch to another tab, then switch back to the Showcase tab

Expected: The Showcase tab should remember your scroll offset
Actual: The Showcase tab is scrolled up to the top again

dfabulich:
Fixes #397

This is underdocumented in https://developer.android.com/guide/navigation/navigation-3/naventrydecorators but there's a bit more documentation in https://developer.android.com/reference/kotlin/androidx/navigation3/ui/NavDisplay.composable and a sample in https://github.com/android/nav3-recipes/tree/main/app/src/main/java/com/example/nav3recipes/multiplestacks

Nav 3 has a `NavDisplay` composable that accepts a `backStack`, and a `NavDisplay` composible that accepts a `List<NavEntry>`. The `backStack` API is a convenience API that calls the `entries` API, like this:

```kotlin
val entries =
    rememberDecoratedNavEntries(
        backStack = backStack,
        entryDecorators = entryDecorators,
        entryProvider = entryProvider,
    )

NavDisplay(
    entries = entries,
    // ...
)
```

Each `NavEntry` gets its own `backStack` with its own decorators, the most important of which is the builtin decorator that provides a `SaveableStateHolder`, which ensures that `remember`ed state gets restored. In Nav 3, you call `rememberDecoratedNavEntries` to associate a `backStack` with its decorators.

We were remembering just one `SaveableStateHolder` decorator for the active `backStack`, but we needed to remember N `SaveableStateHolder` decorators for N tabs.

Here, we're calling `rememberDecoratedNavEntries` 100 times for 100 slots, each with its own `backStack`, its own decorators (just the one `SaveableStateHolder` decorator for each), all sharing the same `tabEntryProvider`.

We pass `NavDisplay` the current `entries` (a list of just one entry) for the currently selected tab. When we switch tabs, we pass `NavDisplay` a different list of one entry, with its own `backStack` and `SaveableStateHolder`; when returning to a previous tab, we provide the remembered entry list, which provides its remembered `SaveableStateHolder`, so the tab can restore its state from that.

Skip Pull Request Checklist:

  (no change required)

-----

-----

### DIFF
--- a/Sources/SkipUI/SkipUI/Containers/TabView.swift
+++ b/Sources/SkipUI/SkipUI/Containers/TabView.swift
@@ -64,6 +64,7 @@ import androidx.compose.ui.unit.dp
 import androidx.navigation3.runtime.NavBackStack
 import androidx.navigation3.runtime.NavEntry
 import androidx.navigation3.runtime.NavKey
+import androidx.navigation3.runtime.rememberDecoratedNavEntries
 import androidx.navigation3.runtime.rememberNavBackStack
 import androidx.navigation3.runtime.rememberSaveableStateHolderNavEntryDecorator
 import androidx.navigation3.ui.NavDisplay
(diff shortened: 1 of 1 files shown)


dfabulich:
@fhasse95 Sorry for the inconvenience! Bug #397 was my fault. I think this PR will fix the bug in Google's recommended way.

marcprux:
I finally had a chance to look at this. It does seem to work for restoring the scroll position for the top-level list in a tab (like the "Showcase" playground list), but then drilling into another view and switching tabs away and then back does not seem to restore the scroll position. For example, the List playground might be scrolled to here:

Screenshot_1776881426

but then switching the the "About" tab and back to the "Showcase" resets the list to the top:

Screenshot_1776881438

PR #393, opened 2026-04-13
Bugfix: Maintain content and scroll offset of the tab view when switching tabs

fhasse95:
This PR resolves an issue in which switching tabs caused the view and scroll positions of the content to reset.

Previously, the tab view worked as expected. However, after updating to the latest version of Skip UI (which migrated to Navigation3), the views began to re-initialize with every tab switch, resulting in the loss of all temporary state and scroll offsets.

This PR ensures that tab views only load once and are then merely toggled. This restores the previous behavior, maintaining the user's position and data when navigating between tabs.

**Before:**
before

**After:**
after

---

Thank you for contributing to the Skip project! Please review the contribution guide at https://skip.dev/docs/contributing/ for advice and guidance on making high-quality PRs.

Use this space to describe your change and add any labels (bug, enhancement, documentation, etc.) to help categorize your contribution.

Skip Pull Request Checklist:

- [X] REQUIRED: I have signed the [Contributor Agreement](https://github.com/skiptools/clabot-config)
- [X] REQUIRED: I have tested my change locally with `swift test`
- [ ] OPTIONAL: I have tested my change on an iOS simulator or device
- [ ] OPTIONAL: I have tested my change on an Android emulator or device
- [X] REQUIRED: I have checked whether this change requires a corresponding update in the [Skip Fuse UI](https://github.com/skiptools/skip-fuse-ui) repository (link related PR if applicable)
=> No Changes Required
- [ ] OPTIONAL: I have added an example of any UI changes in the [Showcase](https://github.com/skiptools/skipapp-showcase-fuse) sample app

-----

- [ ] AI was used to generate or assist with generating this PR. *Please specify below how you used AI to help you, and what steps you have taken to manually verify the changes*.

-----

PR #404, opened 2026-04-29
Add support for `statusBarHidden(:)`

dfabulich:
[Screen_recording_20260429_122620.webm](https://github.com/user-attachments/assets/4d4c13d3-f391-4ba2-bb8b-0b8287d5fdc3)

This PR replaces PR #386. I jumped through a _ton_ of hoops trying to get that to work, but it turned out that all I needed was a magic flag, [`WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE`](https://developer.android.com/reference/androidx/core/view/WindowInsetsControllerCompat#BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE()).

> Option for [setSystemBarsBehavior](https://developer.android.com/reference/androidx/core/view/WindowInsetsControllerCompat#setSystemBarsBehavior(int)): Window would like to remain interactive when hiding navigation bars by calling [hide](https://developer.android.com/reference/androidx/core/view/WindowInsetsControllerCompat#hide(int)) or [setInsetsAndAlpha](https://developer.android.com/reference/androidx/core/view/WindowInsetsAnimationControllerCompat#setInsetsAndAlpha(androidx.core.graphics.Insets,float,float)).
> 
> When system bars are hidden in this mode, they can be revealed temporarily with system gestures, such as swiping from the edge of the screen where the bar is hidden from. These transient system bars will overlay app’s content, may have some degree of transparency, and will automatically hide after a short timeout.

Without this flag, the status bar takes a full second to fade away, and then the window insets change suddenly, creating a jarring layout shift. With this flag, the status bar animates away in 0.5s, and the inset animates with it. The flag also creates this effect where if you swipe down from the top of the screen, the status bar reappears. That seems fine to me.

They also offer a callback to use when it's time to re-hide the status bar, which we're using.

Skip Pull Request Checklist:

- [x] REQUIRED: I have signed the [Contributor Agreement](https://github.com/skiptools/clabot-config)
- [x] REQUIRED: I have tested my change locally with `swift test`
- [ ] OPTIONAL: I have tested my change on an iOS simulator or device
- [x] OPTIONAL: I have tested my change on an Android emulator or device
- [x] REQUIRED: I have checked whether this change requires a corresponding update in the [Skip Fuse UI](https://github.com/skiptools/skip-fuse-ui) repository (link related PR if applicable)
    https://github.com/skiptools/skip-fuse-ui/pull/97
[...]

PR #386, opened 2026-04-04
Add support for `statusBarHidden(:)`

dfabulich:
This one was surprisingly tricky. Partly because the default animation for hiding the status bar is mindbogglingly bad.

Here's how it looked with a naive hide/show implementation: a slow 1-second fade out, followed by a non-animated layout shift.

[Screen_recording_20260403_191618.webm](https://github.com/user-attachments/assets/9a9ed8e2-1d71-4af4-a0d6-c56f0e243651)

And here's how it looks in this PR, with animations disabled. Much better, IMO.

[Screen_recording_20260403_193033.webm](https://github.com/user-attachments/assets/149f398d-b1c0-4c78-9465-9026d5258da5)

The other reason this was surprisingly tricky was that the API for disabling animations seems to be buggy. You have to ask for an animation controller with a callback from [`controlWindowInsetsAnimation`](https://developer.android.com/reference/android/view/WindowInsetsController#controlWindowInsetsAnimation(int,%20long,%20android.view.animation.Interpolator,%20android.os.CancellationSignal,%20android.view.WindowInsetsAnimationControlListener)) , but if you make the call "too early" it simply doesn't work. There's a callback that you're supposed to wait for first, [`onControllableInsetsChanged`](https://developer.android.com/reference/android/view/WindowInsetsController.OnControllableInsetsChangedListener), but _that_ doesn't fire unless you've first waited for the view to be attached with [`View.OnAttachStateChangeListener`](https://developer.android.com/reference/android/view/View.OnAttachStateChangeListener), and even when it _does_ fire, `controlWindowInsetsAnimation` _still_ doesn't work that early.

The _other_ reason it was surprisingly tricky was that the status bar automatically shows back up when the window loses focus (when you open the "recents"/"overview" app-switcher screen with the square button on the emulator). When that happens, the animation controller is cancelled, and you have to request a new one. `SystemOverlays.swift` is 70 lines of extremely hard-earned code.

Skip Pull Request Checklist:

- [x] REQUIRED: I have signed the [Contributor Agreement](https://github.com/skiptools/clabot-config)
- [x] REQUIRED: I have tested my change locally with `swift test`
- [x] OPTIONAL: I have tested my change on an iOS simulator or device
- [x] OPTIONAL: I have tested my change on an Android emulator or device
- [x] REQUIRED: I have checked whether this change requires a corresponding update in the [Skip Fuse UI](https://github.com/skiptools/skip-fuse-ui) repository (link related PR if applicable)
[...]

PR #410, opened 2026-05-01
TabView: Keep a stable SaveableStateHolder for each tab's backStack

### LINKED ISSUE #409: Tab hiding doesn't work
To repro: Launch the Showcase Lite app, navigate to the SafeArea playground --> Geometry padding. There's a button there to Hide/Show the Tab Bar.

But when you tap it to hide the tab bar, the tab bar remains, and doesn't hide.

dfabulich:
* Fixes #409 

Relevant links
* Bug #397
* PR #393
    * PR 393 would have fixed bug 397, but was closed in favor of PR 400
* PR #398
    * PR 398 would have fixed bug 397 differently, but was closed in favor of PR 400
    * If we'd merged it, it would have had "Bug 398x", defined below
* PR #400
    * PR 400 fixed bug 397, but caused Bug 409

This PR looks bigger than it is, but it's actually pretty lightweight, especially if you review it commit-by-commit, looking at it with `git log --first-parent`. 

Understanding it requires recapping recent history a bit.

## "Bug 398x"

PR #398 was my attempt to fix Bug #397 "the right way." It never merged, because @marcprux found a bug in it, which I didn't have time to fix last week:

> It does seem to work for restoring the scroll position for the top-level list in a tab (like the "Showcase" playground list), but then drilling into another view and switching tabs away and then back does not seem to restore the scroll position.

There was never a filed bug filed about that (which makes sense, because PR 398 never actually merged), but I'm calling that bug "398x".

"Bug 398x" wasn't restricted to scroll state. If we'd merged PR #398, _any_ state in views pushed onto the navigation stack would be forgotten when switching tabs. The simple repro for "Bug 398x" is:

* Navigate to the Button playground
* Tap one of the buttons; the Tap Count increases to 1
* Navigate to the About tab and then back to the Showcase tab; the Tap Count is now back down to 0.

**It turns out that Bug 398x had a very simple fix that took me _ages_ to track down.** I just needed to mark the `NavKey`s `SkipNavigationStackRootKey` and `SkipNavigationStackPushKey` as `Hashable`. A two-line fix, but figuring that out was… challenging.

## Problems with PR #400

Bug #409 was caused by PR #400. I think it would be possible to fix #409 building on top of PR #400, but I don't think that's wise.

PR #400's approach was to keep all of the tabs actually rendered, hiding them with alpha/zIndex, like PR #393.

[I had a comment on #393:](https://github.com/skiptools/skip-ui/pull/393#pullrequestreview-4130974414)

> This PR is keeping all of the tabs rendered at once, with inactive tabs transparent and zIndex'd behind the active tab.
> 
> I did a bunch of research into this, and I think this is not Google's recommended way to preserve scroll offset in Navigation 3. I filed a different PR that I think is the "right way."
> 
> * #398

This comment also applies to PR #400.

I think merging PR #400 last week was the right call, that in light of the fact that I wasn't available to fix PR #398. It would have been really bad if Bug #397 had been completely unfixed for an entire week.

But, now that we have some more time, I think we can and should fix this the "right way" here in this PR.

(FWIW, PR #400 caused bug #409 by confusing the preferences collector. It was getting its wires crossed with the other active tabs. My PR #398 never had bug #409 to begin with.)

## What this PR actually does

1. Revert PR 400 in ed093025c0e17f930cc0f9f144109c3b585934bd
2. Merge PR 398 (my "right way" PR) in 5c2decd55cedbe5d73160b3eb362e22401ea9887 (the meat is in the second parent 746b25e1b03ffc43805af254959d2e8371118c53)
3. Fix Bug 398x with a two-line fix in ab7cc9cee6e05c06e4d65fbe16903d8f9f19ce09
4. Fixing this bug created a new bug: `.statusBarHidden(:)` had broken layout; fixed in ffeddca057a3a1836fcc7de3540dfa9f68fe1646

Skip Pull Request Checklist:

    Not required

-----

-----

### DIFF
--- a/Sources/SkipUI/SkipUI/Containers/Navigation.swift
+++ b/Sources/SkipUI/SkipUI/Containers/Navigation.swift
@@ -673,12 +673,12 @@ public struct NavigationStack : View, Renderable {
 #if SKIP
 
 // SKIP INSERT: @Serializable
-public enum SkipNavigationStackRootKey : NavKey {
+public enum SkipNavigationStackRootKey : NavKey, Hashable {
     case root
 }
 
 // SKIP INSERT: @Serializable
-public struct SkipNavigationStackPushKey : NavKey {
+public struct SkipNavigationStackPushKey : NavKey, Hashable {
     public let destinationIndex: Int
     public let identifier: String
 }
--- a/Sources/SkipUI/SkipUI/Containers/TabView.swift
+++ b/Sources/SkipUI/SkipUI/Containers/TabView.swift
@@ -3,6 +3,8 @@
 #if !SKIP_BRIDGE
 import Foundation
 #if SKIP
+import androidx.compose.animation.fadeIn
+import androidx.compose.animation.fadeOut
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.interaction.MutableInteractionSource
@@ -59,9 +61,15 @@ import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.semantics.testTagsAsResourceId
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
-import androidx.compose.ui.zIndex
+import androidx.navigation3.runtime.NavBackStack
+import androidx.navigation3.runtime.NavEntry
+import androidx.navigation3.runtime.NavKey
+import androidx.navigation3.runtime.rememberNavBackStack
+import androidx.navigation3.runtime.rememberSaveableStateHolderNavEntryDecorator
+import androidx.navigation3.ui.NavDisplay
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
+import kotlinx.serialization.Serializable
 #endif
 
 // SKIP @bridge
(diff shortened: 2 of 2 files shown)


dfabulich:
I just added in one more bug fix: this PR was about to break `.statusBarHidden(:)`. I've fixed it in ffeddca057a3a1836fcc7de3540dfa9f68fe1646

The bug happened because we're now using Compose "correctly," reinflating tabs based on their remembered state, but we weren't correctly signaling to Compose that the TabView's content needs to rerender when the `_safeArea` passed into the TabView changes. I added in a call to `rememberUpdatedState`, informing Compose that the tab content depends on the TabView's safeArea; now, the layout works just right, whether the status bar is hidden or not.

<details><summary>Fun fact:</summary>
<p>

The broken `.statusBarHidden(:)` bug was the reason why I initially withdrew my PR #386 on April 4, marking it as a draft. I couldn't figure out what was causing it.

We merged the new PR #404 adding support `.statusBarHidden(:)` which worked just fine (magically, or so it seemed to me) because PR #400 didn't have a bug in `.statusBarHidden(:)`; it kept all the tabs alive, and they all rerendered as the safeArea updated.

But once I dusted off #398 and fixed its state bug, the `statusBarHidden(:)` bug cropped back up. These are the perils of using Compose "correctly": if you don't use it correctly _enough_, it causes other, deeper problems.

</p>
</details>

marcprux:
This works well in my testing. My only regret is reverting to the fixed 100 element state array, but perhaps future work can explore getting rid of that implementation quirk.

PR #418, opened 2026-05-08
Fix crash on state reset when preferences are queried

### LINKED ISSUE #300: NullPointerException in skip.ui.NavigationStack.Render after Android system font size change - causes app to crash
I originally filed this in the wrong place (Skip general discussion), so am closing it there and re-opening here. Sorry for any confusion.

I discovered this when testing my Skip lite app on a physical device (Pixel 6a/Android 16), then reproduced in Emulator, and have found that it’s reproducible in the Skip Showcase app on the Google Play Store. 

The app reproducibly crashes when navigating between tabs with NavigationStack after changing the system font size between extreme sizes. The number of steps to trigger the crash may be dependent on layout complexity. I’ve been able to reproduce it reliably in the Skip Showcase app downloaded from the Google Play Store, which takes an extra few steps compared to my app with a more complex layout.

Excerpt of stacktrace:

```
FATAL EXCEPTION: main
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Object skip.ui.Preference.getReduced$SkipUI_debug()' on a null object reference
    at skip.ui.NavigationStack.Render(Navigation.kt:143)
    at skip.ui.View._ComposeContent(...)
    at showcase.fuse.SettingsView._ComposeContent(...)
    // ... continues into Compose/Android internals
```

Steps to reproduce:
1. Download Skip Showcase app from the Google Play Store
2. Open Android system Settings app, go to "Display & touch" and then "Display size and text"
3. Set Font Size to largest size
4. Open Skip Showcase app
5. Tap Showcase to switch to Showcase tab
6. Switch to system Settings -> Display size and text
7. Set Font Size to smallest size
8. Switch back to Skip Showcase app
9. Tap About to switch to About tab
10. Switch to system Settings -> Display size and text
11. Set Font Size to largest size
12. Switch back to Skip Showcase app
13. Tap Showcase to switch to Showcase tab
 -- This should trigger a crash (if no crash, repeat pattern of changing font size and back to app and switch between tabs)

### LINKED ISSUE #417: NullPointerException in PresentationRoot when app process is killed and recreated (e.g. after revoking permissions from system Settings)
FATAL EXCEPTION: main
Process: com.xxx.app
java.lang.NullPointerException: Attempt to invoke virtual method 
'java.lang.Object skip.ui.Preference.getReduced$SkipUI_debug()' 
on a null object reference
    at skip.ui.PresentationRootKt.PresentationRoot$lambda$1(PresentationRoot.kt:45)
    at skip.ui.PreferenceValues.collectPreferences$SkipUI_debug(PreferenceKey.kt:65)
    at skip.ui.PresentationRootKt.PresentationRoot(PresentationRoot.kt:44)
Steps to Reproduce:
1. Open the app and grant location permission.
2. Go to System Settings → Permissions and revoke location access.
3. Return to the app.
4. The app crashes while attempting to rebuild the composition.

Root Cause:
Android kills the process when the user navigates to System Settings. Upon returning, SaveableStateHolder attempts to restore the state, but Preference.getReduced() returns null, which is not handled by SkipUI.

marcprux:
This consolidates the logic for remembering preferences, and also handled potential null values which can come from the state being reset.

Fixes https://github.com/skiptools/skip-ui/issues/417. Fixes https://github.com/skiptools/skip-ui/issues/300.

### DIFF
--- a/Sources/SkipUI/SkipUI/Containers/Navigation.swift
+++ b/Sources/SkipUI/SkipUI/Containers/Navigation.swift
@@ -134,14 +134,10 @@ public struct NavigationStack : View, Renderable {
 
     #if SKIP
     @Composable public override func Render(context: ComposeContext) {
-        // Have to use rememberSaveable for e.g. a nav stack in each tab
-        let destinations = rememberSaveable(stateSaver: context.stateSaver as! Saver<Preference<NavigationDestinations>, Any>) { mutableStateOf(Preference<NavigationDestinations>(key: NavigationDestinationsPreferenceKey.self))
-        }
-        // Make this collector non-erasable so that destinations defined at e.g. the root nav stack layer don't disappear when you push
-        let destinationsCollector = PreferenceCollector<NavigationDestinations>(key: NavigationDestinationsPreferenceKey.self, state: destinations, isErasable: false)
-        let destinationLayoutHints = rememberSaveable(stateSaver: context.stateSaver as! Saver<Preference<NavigationDestinationLayoutHintsMap>, Any>) { mutableStateOf(Preference<NavigationDestinationLayoutHintsMap>(key: NavigationDestinationLayoutHintsPreferenceKey.self))
-        }
-        let destinationLayoutHintsCollector = PreferenceCollector<NavigationDestinationLayoutHintsMap>(key: NavigationDestinationLayoutHintsPreferenceKey.self, state: destinationLayoutHints, isErasable: false)
+        // Have to use rememberSaveable for e.g. a nav stack in each tab. Make the collectors non-erasable so that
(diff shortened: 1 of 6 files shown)


marcprux:
As mentioned at https://github.com/skiptools/skip-ui/issues/417#issuecomment-4412493161, we don't see this in Showcase due to https://github.com/skiptools/skipapp-showcase/commit/5bb3049be6c883a6d55aa52d52e0f3109a8628ba#diff-e7075a061fc36788a2762feb445a61db436a6f07c529859334189e53914a04f7R53 which adds to Main.kt:

```
SideEffect { saveableStateHolder.removeState(true) }
```
 I wonder if we should be adding that to every project's Main.kt, @aabewhite?

PR #423, opened 2026-05-11
Make Image/AsyncImage stop using subcompose by default

dfabulich:
Coil's documentation strongly recommends against using `SubcomposeAsyncImage` for performance reasons. https://coil-kt.github.io/coil/compose/

Here, we've replaced it with the most flexible version of Coil's non-subcompose API, `rememberAsyncImagePainter`.

But there is an important quirk in this API. SwiftUI `AsyncImage` has three `AsyncImagePhase` cases: `.success(Image)`, and `.failure(Error)`, and `.empty`, but there are four states of `AsyncImagePainter.State`: `Success`, `Failure`, `Loading` and `Empty`.

`Empty` is quite different from `Loading`. In the `Empty` state, which occurs on the first frame of rendering, Coil doesn't know yet whether the image will render immediately from the memory cache. In the subsequent frame, the state will change to either `Success` or `Loading`; it's up to the user to decide what to do in the `Empty` state.

Coil users have a few options for handling `Empty`.

1. We can optimistically render the image, hoping to get a cache hit.
2. We can pessimistically render the placeholder, waiting to render the image until we can be certain it's ready.
3. We can render the placeholder underneath the image, in a ZStack/Box. If the image renders (and if the image doesn't have transparency), the image will completely obscure the placeholder.

For `AsyncImage(url:scale:)` and `AsyncImage(url:scale:content:placeholder:)`, we've chosen option 3 for Coil's `Empty` case.

Users can choose their own option with `AsyncImage(url:scale:content:)`, which accepts an `AsyncImagePhase`. To permit users to distinguish Coil's `Loading` from `Empty`, we've added an argument to SwiftUI's `AsyncImagePhase.empty` enum case; it's now `.empty(Image?)`. This allows users to access the image in the `.empty` case with `let image: Image? = phase.image`. If the image is `nil`, then the image is `Loading`; users can render their placeholder. (In the `.empty` case, `phase.image` will always be `nil` in SwiftUI.) If the image is not `nil`, users can choose what to do with it, probably selecting one of the three options above.

Existing users of `AsyncImage(url:scale:content:)` can continue to handle `case .empty` without changing their code. In practice, that will function like the pessimistic option 2.

We've also introduced a new modifier, `.subcomposeAsyncImage()`, which sets an environment value, causing SkipUI to use `SubcomposeAsyncImage`, the old way.

Skip Pull Request Checklist:

    https://github.com/skiptools/skip-fuse-ui/pull/105
- [x] OPTIONAL: I have added an example of any UI changes in the [Showcase](https://github.com/skiptools/skipapp-showcase-fuse) sample app
    https://github.com/skiptools/skipapp-showcase/pull/99
    https://github.com/skiptools/skipapp-showcase-fuse/pull/67
-----

Cursor generated a first draft; I significantly refactored it. I tested it in Showcase Lite and Fuse. I tested by ensuring that the `empty` branch was hit, then I uncommented `.subcomposeAsyncImage()` (and added temporary logging in `skip-ui` to prove that it actually used subcompose), and verified that the new `empty` branch was not reached in subcompose mode. (The new `empty` branch is never reached in iOS SwiftUI.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2904,6 +2904,132 @@ Image(systemName: "Icons.Filled.Settings")
 #endif
 ```
 
+#### First-frame rendering and `AsyncImagePhase` quirks
+
+SkipUI's `AsyncImage` is built on top of [Coil for Android](https://coil-kt.github.io/coil/), which was inspired by SwiftUI. Just like SwiftUI, Coil allows its users to specify an URL and a placeholder, or to react to phase changes as the image loads.
+
+But there is one major difference between Coil's `AsyncImage` and SwiftUI's `AsyncImage`. On the very first frame, Coil cannot know whether the image will be successfully rendered from the memory cache, because it can't yet know the size of the layout constraints.
+
+This may impact you if you use [`AsyncImage(url:scale:transaction:content:)`](https://developer.apple.com/documentation/swiftui/asyncimage/init(url:scale:transaction:content:)), where the `content` callback accepts an [`AsyncImagePhase`](https://developer.apple.com/documentation/swiftui/asyncimagephase) enum, with three phases:
+
+* `empty`: No image is loaded
+* `failure(any Error)`: An image failed to load with an error
+* `success(Image)`: An image successfully loaded
+
+Instead of three phases, Coil has _four_ phases: `Success`, `Failure`, `Loading`, and `Empty`. `Empty` is the state where Coil doesn't yet know whether the image is ready or not.
+
+To model this, SkipUI's `empty` case is actually `empty(Image?)`. You can use `let image: Image? = phase.image` to read it. If the phase is `empty` and the image is `nil`, then Coil is `Loading`, and you should show your placeholder.
+
+If the image is not `nil` in the `empty` case, you can decide what to do with that image.
+
+There are a few options available to you:
+
+1. Option 1 (Optimistic): You can optimistically render the image, hoping to get a cache hit, delaying rendering the placeholder.
+
+    ```swift
+    AsyncImage(url: url) { phase in
+        switch phase {
+        case .empty:
+            if let image = phase.image {
+                image.resizable()
+            } else {
+                ProgressView()
+            }
+        case .failure:
+            Color.red
+        case .success(let image):
+            image.resizable()
+        }
+    }
+    ```
+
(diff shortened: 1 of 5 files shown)


marcprux (changes requested):
This is currently breaking the ImagePlayground's ["Complex Layout"](https://github.com/skiptools/skipapp-showcase/blob/c42b1dc1da628e9a37521a561b51a0dd768870e3/Sources/Showcase/ImagePlayground.swift#L371), which on iOS looks like:

Image

and prior to this change, it matched on Android:

Image

But with this change applied, the image expands to fill the area:

Image

dfabulich:
I've fixed the Complex Layout Playground in a secondary commit, worth independently reviewing, because this code is kinda gnarly.

The bug was that when a painter has no intrinsic size, `RenderPainter` slaps `fillSize()` on it. I didn't touch `RenderPainter` in this PR, so I don't understand why it uses `fillSize()` on painters with no intrinsic size. Even when the painter recomposes with an intrinsic size, the painter's `fillSize()` had already "claimed" maximum vertical space, and so it never fixes itself on recompose.

I thought about getting rid of `fillSize()` in `RenderPainter`, but I don't want to change something like that without knowing what I'd break.

More broadly, it's not at all clear to me what job RenderPainter and ImageLayout are supposed to be accomplishing (other than tinting). I experimented with switching Complex Layout over to AsyncImage and it just worked, by _not_ trying to apply fillSize to the placeholder, _not_ trying to apply a custom layout. It's just… rendering the image.

And the fact that asset images are using asynchronous loading is not great. https://github.com/skiptools/skip-ui/issues/154 Asset images shouldn't require a loading placeholder! (I've already long ago stopped using SkipUI `Image`, for performance reasons.)

dfabulich:
I believe this PR is the optimal fix for #426.

timkoss:
I tested this with

.package(url: "https://github.com/dfabulich/skip-ui", branch: "nonsubcompose-async-image"),

And it works fine on my Android

marcprux:
This is looking good in my testing. I haven't checked all the edges, but I'll go ahead and merge this and we can keep an eye out for image quirks going forward.

> And the fact that asset images are using asynchronous loading is not great.

Completely agree. As I mentioned back in https://github.com/skiptools/skip-ui/issues/154#issuecomment-2992350676, I think the solution might be to switch to some other image loading library, but I am loathe to have two separate image libraries (with different capabilities, quirks, and image format support) running simultaneously, where a bundled asset image might be rendered differently from an image downloaded from the web.

PR #424, opened 2026-05-12
Fix stale gradient brush cached at composition

marcprux:
Rather than remembering the individual parameters going into the brush, this PR has us just remember the `LinearGradientShaderBrush`/`RadialGradientShaderBrush` itself. Just remembering the parameters could result in stale cached brushes being used.


marcprux:
The issue this aims to solve is when there were many gradients being rendered simultaneously, such as in the [Sirtet game](https://github.com/Faire-Games/Faire-Games/blob/main/Sources/Tetris/Tetris.swift#L754), the preview blocks (which should be just a single color) would accidentally re-use other cached gradients.

Screenshot_1779193868

PR #441, opened 2026-05-19
Revert "Fix crash on state reset when preferences are queried"

### LINKED ISSUE #300: NullPointerException in skip.ui.NavigationStack.Render after Android system font size change - causes app to crash

### LINKED ISSUE #440: `NavigationStackPlayground` infinitely recomposes
The Showcase `NavigationStackPlayground` infinitely recomposes.

dfabulich:
This reverts commit 23cf094ecec5d142e11034b1100590329e86ee62 (PR #418).

Fixes #440

This PR may feel a bit trollish, but, I think it might actually be the right thing to do.

#418 caused #440 by reading preference state (thus declaring a dependency on that state) while constructing a nav entry, triggering a recomposition infinite loop. I've only come up with one good way to accomplish what #418 accomplishes without reading the state… and that's to clean up the state in `Main.kt`.

Remember, the reason we couldn't reproduce #300 or #417 is that Showcase's `Main.kt` had `SideEffect { saveableStateHolder.removeState(true) }` in it. In #418, @marcprux commented:

> As mentioned at [#417 (comment)](https://github.com/skiptools/skip-ui/issues/417#issuecomment-4412493161), we don't see this in Showcase due to [skiptools/skipapp-showcase@5bb3049#diff-e7075a061fc36788a2762feb445a61db436a6f07c529859334189e53914a04f7R53](https://github.com/skiptools/skipapp-showcase/commit/5bb3049be6c883a6d55aa52d52e0f3109a8628ba#diff-e7075a061fc36788a2762feb445a61db436a6f07c529859334189e53914a04f7R53) which adds to Main.kt:
> 
> ```
> SideEffect { saveableStateHolder.removeState(true) }
> ```
> 
> I wonder if we should be adding that to every project's Main.kt, @aabewhite?

We already _do_ add that to every project's `Main.kt`. Abe updated the project template to add that line in October 2024. https://github.com/skiptools/skipstone/commit/f53a2ced5250e84dbd61168162842afed497b78d It shipped as part of skip 1.7.0.

PR #418 is a heroic attempt to fix #300 / #417 without requiring users to update their `Main.kt` file. But, since that causes #440, I think users should just update their `Main.kt` files, instead. (Maybe we could/should add a `skip verify` step to check for this or something.)

To be clear, here are my test results:

1. When I run my 300 repro test on main skip-ui and main Showcase, it passes.
1. When I revert 418 in `skip-ui` and run my 300 repro test on main Showcase, it still passes, because of `SideEffect { saveableStateHolder.removeState(true) }` in `Main.kt`.
1. When I run my 300 repro test on main skip-ui and remove `SideEffect { saveableStateHolder.removeState(true) }` from Showcase's `Main.kt`, it still passes, because 418 is doing its job.
1. Only when I _both_ revert 418 _and_ remove `SideEffect { saveableStateHolder.removeState(true) }`, _then_ my 300 repro test fails.

Skip Pull Request Checklist:

    Not required

-----

-----

### DIFF
--- a/Sources/SkipUI/SkipUI/Containers/Navigation.swift
+++ b/Sources/SkipUI/SkipUI/Containers/Navigation.swift
@@ -137,10 +137,14 @@ public struct NavigationStack : View, Renderable {
 
     #if SKIP
     @Composable public override func Render(context: ComposeContext) {
-        // Have to use rememberSaveable for e.g. a nav stack in each tab. Make the collectors non-erasable so that
-        // destinations defined at e.g. the root nav stack layer don't disappear when you push.
-        let (destinations, destinationsCollector) = rememberSaveablePreferenceCollector(key: NavigationDestinationsPreferenceKey.self, stateSaver: context.stateSaver as! Saver<Preference<NavigationDestinations>, Any>, isErasable: false)
-        let (destinationLayoutHints, destinationLayoutHintsCollector) = rememberSaveablePreferenceCollector(key: NavigationDestinationLayoutHintsPreferenceKey.self, stateSaver: context.stateSaver as! Saver<Preference<NavigationDestinationLayoutHintsMap>, Any>, isErasable: false)
+        // Have to use rememberSaveable for e.g. a nav stack in each tab
+        let destinations = rememberSaveable(stateSaver: context.stateSaver as! Saver<Preference<NavigationDestinations>, Any>) { mutableStateOf(Preference<NavigationDestinations>(key: NavigationDestinationsPreferenceKey.self))
+        }
+        // Make this collector non-erasable so that destinations defined at e.g. the root nav stack layer don't disappear when you push
+        let destinationsCollector = PreferenceCollector<NavigationDestinations>(key: NavigationDestinationsPreferenceKey.self, state: destinations, isErasable: false)
+        let destinationLayoutHints = rememberSaveable(stateSaver: context.stateSaver as! Saver<Preference<NavigationDestinationLayoutHintsMap>, Any>) { mutableStateOf(Preference<NavigationDestinationLayoutHintsMap>(key: NavigationDestinationLayoutHintsPreferenceKey.self))
+        }
+        let destinationLayoutHintsCollector = PreferenceCollector<NavigationDestinationLayoutHintsMap>(key: NavigationDestinationLayoutHintsPreferenceKey.self, state: destinationLayoutHints, isErasable: false)
         let reducedDestinations = destinations.value.reduced
         let reducedDestinationLayoutHints = destinationLayoutHints.value.reduced
         let mergedDestinations = mergeNavigationDestinationsWithLayoutHints(reducedDestinations, layoutHints: reducedDestinationLayoutHints)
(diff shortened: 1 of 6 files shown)


marcprux:
This looks good in my testing, and hopefully it will fix https://github.com/skiptools/skip-ui/issues/462.

Thanks for the fix, and apologies for the delay reviewing and merging!

marcprux:
In hindsight, could this have been solved more simply with just this `REMOVEME`?

```swift
@Composable func rememberSaveablePreference<V>(stateSaver: Saver<Preference<V>, Any>, initial: () -> Preference<V>) -> MutableState<Preference<V>> {
    let state = rememberSaveable(stateSaver: stateSaver) { mutableStateOf(initial()) }
    // REMOVEME:
    //if (state.value as Any?) == nil {
    //    state.value = initial()
    //}
    return state
}
```

I ask because the `rememberSaveablePreferenceCollector` function was a useful shorthand we used in various places, so it might be worth restoring (but without the special handling of null values).

dfabulich:
I'm not sure I'm reading your example correctly; in particular, your use of `REMOVEME` is not an idiom I'm familiar with.

I think you're saying something like: we could bring back most of #418, but tweak its implementation by deleting just those three lines of code…?

marcprux:
Yes, that's exactly what I was trying to say. In other words, retain the simplicity benefits of the single aggregate `rememberSaveablePreference` function, just so long as it doesn't alter the behavior.

PR #414, opened 2026-05-07
Animate navigation bar visibility

dfabulich:
There are three kinds of changes here.

1. We have to plumb the `visibilityAnimation` through a preference so `Navigation.swift` can access it.
2. Adding in `AnimatedVisibility`. (This changes indentation on the whole file. Best viewed ignoring whitespace, e.g. with `git diff -w`)
3. Careful coordination of `contentSafeArea` with `topPadding`. These need to match each other on every frame of the animation, or the animation will judder.

# Videos

<details><summary>Videos</summary>
<p>

https://github.com/user-attachments/assets/061a8dbd-f9fe-4d7f-b4af-392e3ef422f0

https://github.com/user-attachments/assets/7b30fdc5-2bcc-42e5-8251-c0e01571bb1c

[Screen_recording_20260507_164120.webm](https://github.com/user-attachments/assets/3ace448b-483d-4d03-b4d8-2b26f2009aba)

[Screen_recording_20260507_164143.webm](https://github.com/user-attachments/assets/72d8a13c-1184-4b54-85ae-59e79255378e)

</p>
</details> 

Skip Pull Request Checklist:

- [x] REQUIRED: I have signed the [Contributor Agreement](https://github.com/skiptools/clabot-config)
- [x] REQUIRED: I have tested my change locally with `swift test`
- [ ] OPTIONAL: I have tested my change on an iOS simulator or device
- [x] OPTIONAL: I have tested my change on an Android emulator or device
- [x] REQUIRED: I have checked whether this change requires a corresponding update in the [Skip Fuse UI](https://github.com/skiptools/skip-fuse-ui) repository (link related PR if applicable)
    None required
- [x] OPTIONAL: I have added an example of any UI changes in the [Showcase](https://github.com/skiptools/skipapp-showcase-fuse) sample app
    https://github.com/skiptools/skipapp-showcase/pull/97
-----

- [x] AI was used to generate or assist with generating this PR. *Please specify below how you used AI to help you, and what steps you have taken to manually verify the changes*.

Cursor added in `AnimatedVisibility`; I wrote the code to coordinate `contentSafeArea` and `topPadding`, and tested it in Showcase Lite, in https://github.com/skiptools/skipapp-showcase/pull/97 in both the "Geometry Padding" and "Geometry padding in sheet" playgrounds.

PR #452, opened 2026-06-01
fix(NavigationStack): don't reserve top-bar inset for title-less roots

### LINKED ISSUE #451: NavigationStack reserves a phantom top inset for title-less roots on Android (regression from #414)
**Symptom:** On Android, a `NavigationStack` whose root has no navigation title or toolbar renders its content pushed down by a phantom top inset of ~`safeAreaTop + 112dp` (the height of a large top app bar that is never actually shown). Every title-less root is affected — e.g. tab roots inside a `TabView` show a large empty gap above their content.

**Regression from:** #414 (Animate navigation bar visibility, 624f9006).

**Root cause:** In `Containers/Navigation.swift`, `topBarBottomPx` is initialized to `safeAreaTopPx + defaultTopBarHeight` (112dp) and insets the content (`.insetting(.top, to: topBarBottomPx.value)`). #414 moved the "reset to 0 when hidden" into a `DisposableEffect { onDispose { topBarBottomPx.value = 0 } }` placed *inside* `AnimatedVisibility(visible: showTopBar)`. For a title-less root, `showTopBar` is `false` from the start (`!isRoot || hasTitle || hasItems` → false), so `AnimatedVisibility` never composes its content and the `onDispose` never runs — leaving the inset at its initial 112dp + safe area.

**Repro:** a `NavigationStack { Text("hi") }` with no `.navigationTitle` (e.g. as a `TabView` tab root) on Android → content sits ~safeArea+112dp below the top.

**Fix:** initialize the inset to 0 when `!showTopBar`, plus a `LaunchedEffect(showTopBar)` to clear it on visible→hidden transitions; preserves the 112dp anti-flash init for bars that *are* shown. PR to follow.

vincentborko:
Fixes #451. Regression from #414: the reset of `topBarBottomPx`/`topBarHeightPx` to 0 lives in `onDispose` inside `AnimatedVisibility(visible: showTopBar)`, which never composes (so never disposes) when `showTopBar` is `false` from the start — leaving a phantom ~`safeArea + 112dp` top inset on every title-less root (e.g. `TabView` tab roots with no `.navigationTitle`).

Initialize both to `0` when `!showTopBar`, and add a `LaunchedEffect(showTopBar)` to clear them on visible→hidden transitions. The 112dp anti-flash init is preserved for bars that are shown; pushed/detail screens (`showTopBar == true`) are unaffected.

Verified on Android (Skip Lite, a production app): the gap above title-less tab roots is gone and detail-screen bars/back buttons still render correctly.

### DIFF
--- a/Sources/SkipUI/SkipUI/Containers/Navigation.swift
+++ b/Sources/SkipUI/SkipUI/Containers/Navigation.swift
@@ -305,12 +305,24 @@ public struct NavigationStack : View, Renderable {
         let defaultTopBarHeight = 112.dp
         let topBarBottomPx = remember {
             // Default our initial value to the expected value, which helps avoid visual artifacts as we measure actual values and
-            // recompose with adjusted layouts
+            // recompose with adjusted layouts. Only reserve the inset when a bar will actually be
+            // shown: a title-less root has showTopBar == false and never composes the
+            // AnimatedVisibility content below, so its onDispose reset never runs — initializing to
+            // the bar height here would leave a phantom top inset (~safeArea + 112dp).
             let safeAreaTopPx = arguments.safeArea?.safeBoundsPx.top ?? Float(0.0)
-            mutableStateOf(with(density) { safeAreaTopPx + defaultTopBarHeight.toPx() })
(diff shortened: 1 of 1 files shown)


marcprux:
Funny, I had the exact same fix locally that I just hadn't make a PR for yet 😀

PR #453, opened 2026-06-02
Text + Text concatenation via Compose AnnotatedString

vincentborko:
## What

Implements `Text + Text` concatenation with per-segment styling on Android, rendered as a single Compose `AnnotatedString`.

`static func + (Text, Text)` was previously `fatalError()`, so any SwiftUI that builds a multi-style inline string by concatenation (per-segment color / weight / size / italic / monospaced / underline / strikethrough / gradient foreground) could not run on Android.

## How

Both runtimes feed **one** ordered `[TextRun]` model, folded into a single `AnnotatedString` (each run's styling as a `SpanStyle`) and rendered by the shared `_Text.Render` path — so environment-level concerns (text alignment, line limit + truncation, tracking, line spacing, `material3Text`) behave like a normal `Text`, with **no duplicated render logic**:

- **Skip Lite (transpiled)** — a native `+` operator (`Text.plus`, `// SKIP DECLARE: operator fun plus`). A styled `Text` applies its style as an environment modifier that can't be read back at render time, so each operand's styling is captured as `TextRunStyle` *data* where the modifier is applied (`.foregroundColor`, `.font`, `.bold`, …), mirroring SkipSwiftUI's `TextRun`.
- **SkipFuse** — `// SKIP @bridge public init(bridgedRuns:colors:fontSizes:fontWeights:flags:)` accepts primitive-only per-run descriptors and folds them into the same `[TextRun]`. The foreground resolves to a solid color when it is one, otherwise a Compose brush (gradients via `SpanStyle(brush:)`). `localizedTextString()`, `asColor`, and `asBrush` are `@Composable`, so they're resolved in the `@Composable` scope before the (non-composable) builder lambda.

> **Note (supersedes the original draft):** the native Skip Lite `+` operator *is* implemented here now. Earlier this PR left `+` unimplemented and reached the run model only via the bridge; it has since been unified so `Text + Text` works in **both** Skip Lite and SkipFuse through the one `_Text`-with-runs path.

## Dependencies / pairing

- **Requires** skiptools/skipstone#255 — the transpiler fix that composes a bare, statement-level `Text + Text`. Without it, an unconstrained concatenation (e.g. directly in a `VStack`) is constructed but never composed (zero size); it only renders when wrapped in another view modifier like `.frame(...)`.
- **Pairs with** skiptools/skip-fuse-ui#118 — the SkipSwiftUI side that captures styling as data and calls the `bridgedRuns` initializer.

## Testing

Verified on an Android emulator (Showcase "Text Concatenation" playground): per-run colors, weights, sizes, italic/monospaced, underline/strikethrough, gradient foreground, and multi-line wrapping all render as a single flowing styled block — both bare/unconstrained and width-constrained.

---

- [x] AI was used to assist with this PR.

### DIFF
--- a/Sources/SkipUI/SkipUI/Text/Text.swift
+++ b/Sources/SkipUI/SkipUI/Text/Text.swift
@@ -52,6 +52,15 @@ public struct Text: View, Renderable, Equatable {
     private let textView: _Text
     private let modifiedView: any View
 
+    #if SKIP
+    // Styling captured as *data* (alongside `modifiedView`) so that `Text + Text` concatenation can
+    // rebuild each operand's style as a Compose `SpanStyle`. The styling modifiers (`.foregroundColor`,
+    // `.font`, `.bold`, …) wrap `modifiedView` — used to render a standalone styled `Text` — *and*
+    // record into `capturedStyle`, because a styled `Text` applies its style as an environment modifier
+    // that cannot be read back out at render time to build per-run spans. Mirrors SkipSwiftUI's `Text`.
+    var capturedStyle = TextRunStyle()
+    #endif
+
     // SKIP @bridge
     public init(verbatim: String) {
         textView = _Text(verbatim: verbatim)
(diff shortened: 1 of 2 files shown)


marcprux:
We might consider merging this work with https://github.com/skiptools/skip-ui/pull/434 to support the `+` operator in Skip Lite as well as Fuse.

vincentborko:
Nice idea, I'll update with the best of both worlds shortly! ;)

vincentborko:
all working cleanly now 👍

Sources/SkipUI/SkipUI/Text/Text.swift
```diff
@@ -195,27 +275,51 @@ public struct Text: View, Renderable, Equatable {
     // Text-specific implementations of View modifiers
 
     public func accessibilityLabel(_ label: Text) -> Text {
+        #if SKIP
```
marcprux, line 278: All of these `#if SKIP` additions aren't necessary: the `#else` condition if never used, since this package is never compiled or used in a non-`SKIP` context. You should be able to just change everything to use the `return styled(…)` rather than gating them within the `#if SKIP` blocks.

marcprux (changes requested):
This looks good overall, but I'd love to see one or two test cases in SkipUI to demonstrate the concatenation.

vincentborko:
Added `testTextConcatenationLaysOutAsSingleString` and `testTextConcatenationAppliesPerSegmentFont` in `Tests/SkipUITests/TextTests.swift`. Each fails only for its own reason: dropping the right operand's runs in `plus()` breaks the first, dropping `capturedStyle` in `flattenedRuns` breaks the second. `swift test` is green otherwise (62 tests, 0 failures), with `:SkipUI:compileDebugKotlin` succeeding.

Both compare rendered pixmaps, so they carry the same guard as `testDrawMessage` and `testDrawTextMonospacedFont` in this file:

```swift
if isAndroid {
    throw XCTSkip("Disabled on Android due to inconsistent font rendering on different emulators")
}
```

Without it they failed the connected emulator step on `stable_default_28`, drawn height 22 against a reference of 23 at the correct font size. `isAndroid` is false on the JVM, so Robolectric still runs both in full, which is where the mutations above were measured.

Two known limits, neither addressed here. With an explicit per-run font the concatenation comes out a couple of pixels narrower than the equivalent plain `Text`, so the second test compares drawn height rather than the whole pixmap; native SwiftUI renders the two identically, so `Font.system(size:)` in the `SpanStyle` seems to resolve to slightly different metrics than the environment font does. And `flattenedRuns` returns an existing `runs` array without folding in the outer `capturedStyle`, so `(Text("A") + Text("B")).bold() + Text("C")` loses the bold on the first two segments. Happy to take either one separately.

vincentborko:
Correction to the note above: as first pushed, both cases failed the connected emulator step, and I have amended them to skip there.

They compare rendered pixmaps, so they run into the same per-emulator font rasterisation differences that `testDrawMessage` and `testDrawTextMonospacedFont` already opt out of in this file — on `stable_default_28` the drawn height came out one row short of the reference (`22 != 23`) while rendering at the correct size. Both now carry the same guard and the same wording as the existing cases:

```swift
if isAndroid {
    throw XCTSkip("Disabled on Android due to inconsistent font rendering on different emulators")
}
```

So, precisely: the two cases execute natively and under Robolectric, and are skipped on a connected Android device or emulator. `isAndroid` is false on the JVM, so the Robolectric run still executes them in full — verified in the gradle report (`tests=15 failures=0 errors=0 skipped=0`, both cases with non-zero runtime), and the mutation results in the previous comment were measured there and still hold.

Everything else in that run was green on the emulator, including all eleven existing `testTextSize*` cases against the merge of this branch with current main, so the failure was specific to these two.

Sources/SkipUI/SkipUI/Text/Text.swift
vincentborko: Done in 501b4c8: 14 of the 15 blocks are gone. The one at `Text.swift:105` stays, because `init(bridgedRuns:)` seeds from `_Text.plainTextSeed`, which reads `attributedString.string`, a field only SkipFoundation's `AttributedString` has. `swift test` green, `compileDebugKotlin` succeeds.

PR #455, opened 2026-06-02
Implement presentationDragIndicator(_:)

vincentborko:
## What

Implements `presentationDragIndicator(_:)`, which is currently `@available(*, unavailable)`.

## Why

The bottom-sheet presentation (`SheetPresentation`) already draws its own grab-handle capsule for non-fullscreen sheets, so `.visible` / `.automatic` happen to look correct today. But because the modifier is unavailable, `.hidden` has no effect — there is no way to suppress the handle on Android, which diverges from SwiftUI on iOS.

## How

- Add a collected `PresentationDragIndicatorPreferences` (a `Visibility`), mirroring the existing `PresentationDetentPreferences` machinery, and read its reduced value in `SheetPresentation` to gate the capsule.
- When `.hidden`, the capsule is replaced by a spacer of identical height so the sheet content does not shift.
- Replace the unavailable stub with a real implementation, plus a `// SKIP @bridge` `presentationDragIndicator(bridgedVisibility:)` variant taking the `Visibility` raw value — matching the existing `confirmationDialog(bridgedTitleVisibility:)` bridging pattern.

Single-file change in `Layout/Presentation.swift`.

## Verification

Built and exercised on an Android emulator from a SkipFuse app via a small debug gallery presenting medium/large-detent sheets:
- `.visible` and `.automatic` → grab-handle capsule shown.
- `.hidden` → no capsule, content stays in place.

## Paired PR

Forwarding the bridge from the SkipSwiftUI layer: skiptools/skip-fuse-ui#120


marcprux:
Very nice, thanks!

PR #457, opened 2026-06-03
Simultaneous gesture support

tifroz:
## Summary

- Add partial Android support for `View.simultaneousGesture` so tap gestures and drag observers can run alongside an existing gesture instead of replacing it.
- Add `GestureMask.none` handling so a simultaneous gesture can be explicitly disabled.
- Add an Android nested-scroll helper for pull-down over scroll handoff (i.e. convert a drag-to-scroll into drag-to-pull on over scroll )

## API Shape

```swift
extension View {
    public func simultaneousGesture<V>(_ gesture: any Gesture<V>, isEnabled: Bool) -> some View
    public func bridgedSimultaneousGesture(_ gesture: Any, isEnabled: Bool) -> any View
    public func androidVerticalOverscrollPullDown(
        isEnabled: Bool,
        onPull: @escaping (CGFloat) -> Void,
        onEnd: @escaping () -> Void
    ) -> any View
}
```

## Usage Example

```swift
ScrollView {
    content
}
.simultaneousGesture(
    DragGesture().onChanged { _ in
        observedDragCount += 1
    }
)
```

## Limitations

- This is partial parity. The branch handles the gesture-observer cases covered by tests; `GestureMask.all`, `.gesture`, and `.subviews` are not yet full SwiftUI parity.
- Full `swift test` currently fails in unrelated Android snapshot rendering tests (`CanvasTests.testZStackMultiOpacityOverlay`, `LayoutTests.testRotatedSquareAliased`, `LayoutTests.testRotatedSquare`), so PR validation used focused simultaneous-gesture Robolectric tests.

## Testing

- `swift test --filter testSimultaneous`
- Robolectric targeted gesture tests:
  `gradle testDebug --project-dir .build/plugins/outputs/skip-ui/SkipUITests/destination/skipstone --tests "skip.ui.SkipUITests.testSimultaneous*"`
- Sandbox app testing on device (see android screen capture, description below)

**Active horizontal rail** swipe the tile rail sideways, then drag vertically on it. The rail should still scroll horizontally, while the simultaneous drag observer updates the Active changed, Active ended, translation counter, and the small handle position.

**Vertical scroll observer** drag the vertical row list. The list should keep scrolling normally while the simultaneous drag observer updates the Scroll changed, Scroll ended, and translation counters.

**.none gesture mask rail** drag the second tile rail. It uses the same drag observer with including: .none, so the rail can still be interacted with, but the .none counters should stay at zero.

https://github.com/user-attachments/assets/317942af-0ff0-4688-b884-3dd34756a1a0

### DIFF
--- a/Sources/SkipUI/SkipUI/System/Gesture.swift
+++ b/Sources/SkipUI/SkipUI/System/Gesture.swift
@@ -967,7 +967,7 @@ func detectDragGesturesWithScrollAxes(onDragEnd: () -> Void, onDragCancel: () ->
 }
 
 // SKIP DECLARE: suspend fun PointerInputScope.detectSimultaneousDragGestures(onDrag: (PointerInputChange, Offset) -> Unit, onDragEnd: () -> Unit, onDragCancel: () -> Unit, shouldAwaitTouchSlop: () -> Boolean)
-func detectSimultaneousDragGestures(onDragEnd: () -> Void, onDragCancel: () -> Void, onDrag: (PointerInputChange, Offset) -> Void, shouldAwaitTouchSlop: () -> Bool) {
+func detectSimultaneousDragGestures(onDrag: (PointerInputChange, Offset) -> Void, onDragEnd: () -> Void, onDragCancel: () -> Void, shouldAwaitTouchSlop: () -> Bool) {
     awaitEachGesture {
         let down = awaitFirstDown(requireUnconsumed: false, pass: PointerEventPass.Initial)
         let awaitTouchSlop = shouldAwaitTouchSlop()
--- a/Sources/SkipUI/SkipUI/System/Gesture.swift
+++ b/Sources/SkipUI/SkipUI/System/Gesture.swift
@@ -18,6 +18,7 @@ import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.input.pointer.PointerEventPass
 import androidx.compose.ui.input.pointer.PointerInputChange
 import androidx.compose.ui.input.pointer.PointerInputScope
+import androidx.compose.ui.input.pointer.changedToUpIgnoreConsumed
 import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.input.pointer.positionChange
 import androidx.compose.ui.layout.LayoutCoordinates
@@ -430,7 +431,7 @@ extension View {
 
     public func gesture<V>(_ gesture: any Gesture<V>, isEnabled: Bool) -> any View {
         #if SKIP
-        return ModifiedContent(content: self, modifier: GestureModifier(gesture: gesture as! Gesture<Any>, isEnabled: isEnabled))
+        return ModifiedContent(content: self, modifier: GestureModifier(gesture: gesture as! Gesture<Any>, isEnabled: isEnabled, composition: .normal))
         #else
         return self
         #endif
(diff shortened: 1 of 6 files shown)


Sources/SkipUI/SkipUI/System/Gesture.swift
```diff
@@ -860,6 +965,56 @@ func detectDragGesturesWithScrollAxes(onDragEnd: () -> Void, onDragCancel: () ->
         }
     }
 }
+
+// SKIP DECLARE: suspend fun PointerInputScope.detectSimultaneousDragGestures(onDrag: (PointerInputChange, Offset) -> Unit, onDragEnd: () -> Unit, onDragCancel: () -> Unit, shouldAwaitTouchSlop: () -> Boolean)
+func detectSimultaneousDragGestures(onDragEnd: () -> Void, onDragCancel: () -> Void, onDrag: (PointerInputChange, Offset) -> Void, shouldAwaitTouchSlop: () -> Bool) {
```
marcprux: This function doesn't look like it should need "SKIP DECLARE". Can you explain why it was added? There's also an oddity with it: the ordering of the SKIP DECLARE parameters are `onDrag:onDragEnd:onDragCancel:shouldAwaitTouchSlop:`, but the order of the actual function is `onDragEnd:onDragCancel:onDrag:shouldAwaitTouchSlop:`.

marcprux (changes requested):

Sources/SkipUI/SkipUI/System/Gesture.swift
tifroz: `//SKIP DECLARE` is needed because the generated Kotlin must be a `suspend fun PointerInputScope...;` without it, Skip emits a plain function, so `awaitEachGesture`, `awaitFirstDown`, `viewConfiguration`, etc. lose their receiver context.

I will resubmit after fixing the parameters order

marcprux: Couldn't you just declare `detectSimultaneousDragGestures` `async` in order to generate the `suspend fun`?

tifroz: That's cleaner but after I made the change to `async`, the drag observer Android unit test fails.

I am also investigating a new unrelated issue that I think may have been introduced by a recent merge? `Unresolved reference 'rememberSaveablePreferenceCollector'`

I need to spend a little more time sorting it out, but I should be posting an update tomorrow

marcprux: > I am also investigating a new unrelated issue that I think may have been introduced by a recent merge? Unresolved reference 'rememberSaveablePreferenceCollector'

Yes, this got broken from a PR collision. It should be fixed in https://github.com/skiptools/skip-ui/pull/464.

tifroz: Submitting a new version with a fix of the unrelated `Unresolved reference 'rememberSaveablePreferenceCollector'` (can't build without the fix) + brief comments to explain `//SKIP DECLARE` Vs `async`.

I kept the `//SKIP DECLARE` in place because it is slightly more robust and can be unit-tested on roboelectric while the async version cannot. This is very much a tradeoff decision, I am happy using `async` and removing the unit test if that's your preference

Here is my current understanding:

....with `// SKIP DECLARE` we get
```
suspend fun PointerInputScope.detectSimultaneousDragGestures(...) {
    awaitEachGesture {
        ...
    }
}
```

kotlin generated from `async`:
```kotlin
internal suspend fun PointerInputScope.detectSimultaneousDragGestures(...): Unit = Async.run {
    awaitEachGesture {
        ...
    }
}
```

Notice the `Async.run` wrapper, the main difference between the 2 flavors

This is the failing unit test (With the async flavor, the test failure is `dragChangeCount.value is still 0`):
```swift
rule.waitForIdle()
rule.onNodeWithTag("simultaneous.scroll.content").performTouchInput { swipeUp() }
rule.waitForIdle()
XCTAssertGreaterThan(dragChangeCount.value, 0)
```

In a sandbox SkipFuse app running on a real device, `dragChangeCount.value` does reflect the changed values and everything is working as designed. In the roboelectric unit test, the value is never updated even after I tried inserting more delays to simulate a human interaction.

I am not entirely sure why the behaviors diverge (possible Compose / Bridging ?), but going with `async` would essentially mean we would lose the ability to detect regressions via unit-tests (if there is a smarter way to unit test, I could not figure it out)

marcprux: I agree that keeping it testable is worth the "SKIP DECLARE" ugliness. There's probably some way we could get the Robolectric test to work (e.g., with some explicit mechanism to wait for the await call), but I won't hold up this PR for it.

PR #464, opened 2026-06-17
Inline drag indicator preference collector to handle removal of rememberSaveablePreferenceCollector

marcprux:
Fixed collision introduced by the removal of `rememberSaveablePreferenceCollector` in https://github.com/skiptools/skip-ui/pull/441#issuecomment-4730760120 and the referencing of the function in https://github.com/skiptools/skip-ui/pull/455.


marcprux:
Included in https://github.com/skiptools/skip-ui/pull/457

PR #468, opened 2026-06-22
fix: parse hexadecimal and 8-bit .colorset color components (#146)

### LINKED ISSUE #146: Cannot parse custom colors whose .colorset uses Hex values
Skip's named color handling parses the .colorset resource, but only handles when the "Input Method" is "Floating Point". For example, this setting:

Image

Will create this `Sources/Showcase/Resources/Module.xcassets/CustomRed.colorset/Contents.json` file:

```json
{
  "colors" : [
    {
      "color" : {
        "color-space" : "srgb",
        "components" : {
          "alpha" : "1.000",
          "blue" : "0.733",
          "green" : "0.067",
          "red" : "0.580"
        }
      },
      "idiom" : "universal"
    },
    {
      "appearances" : [
        {
          "appearance" : "luminosity",
          "value" : "dark"
        }
      ],
      "color" : {
        "color-space" : "srgb",
        "components" : {
          "alpha" : "1.000",
          "blue" : "0.474",
          "green" : "0.493",
          "red" : "1.000"
        }
      },
      "idiom" : "universal"
    }
  ],
  "info" : {
    "author" : "xcode",
    "version" : 1
  }
}
```

But if you switch the "Input Method" to "8-bit hexadecimal":

Image

It will create the following contents with hex encoding:

```json
{
  "colors" : [
    {
      "color" : {
        "color-space" : "srgb",
        "components" : {
          "alpha" : "1.000",
          "blue" : "0xBB",
          "green" : "0x11",
          "red" : "0x94"
        }
      },
      "idiom" : "universal"
    },
    {
      "appearances" : [
        {
          "appearance" : "luminosity",
          "value" : "dark"
        }
      ],
      "color" : {
        "color-space" : "srgb",
        "components" : {
          "alpha" : "1.000",
          "blue" : "0.474",
          "green" : "0.493",
          "red" : "1.000"
        }
      },
      "idiom" : "universal"
    }
  ],
  "info" : {
    "author" : "xcode",
    "version" : 1
  }
}
```

Skip's color parsing does not handle this: https://github.com/skiptools/skip-ui/blob/63bc24afa860656c34dba4b90146680c8d3aa74b/Sources/SkipUI/SkipUI/Color/Color.swift#L603-L616

This results in the color just appearing black. We should handle all the Input Method variants that can be represented in the `.colorset` file.

arulagarwal:
## Why

Xcode's Asset Catalog lets each `.colorset` channel use a different "Input Method". SkipUI's Android colorset parser assumed the **Floating Point** method, where each channel is a 0–1 decimal string like `"0.945"`. When a color is authored with the **8-bit Hexadecimal** method, Xcode emits the *same* JSON shape but encodes each channel as a hex byte string like `"0xF1"`. On the transpiled (Android) path these are read with `toDoubleOrNull`, which returns `null`, so every channel falls back to `0.0` and **the color renders pure black** (#146). It is not a crash — JSON decoding succeeds and nothing is logged; the only symptom is wrong pixels.

This affects Android only: on Apple platforms `Color(_:bundle:)` defers to SwiftUI's own asset loader, so the defect lives entirely in the `#if SKIP` colorset code in `Color.swift`.

## What

`ColorComponents` now routes each channel (red/green/blue/alpha) through a new file-level `parseColorComponent(_:defaultValue:)` helper that covers all three of Xcode's numeric Input Methods: an `0x`/`0X`/`#` prefix is parsed as an 8-bit hex byte (via Kotlin's `toIntOrNull(radix:)`, since Swift's `Int(_:radix:)` does not transpile); a bare integer with no decimal point is treated as an 8-bit (0–255) value; anything else is parsed as a floating-point string, so existing decimal colorsets are byte-for-byte unchanged. Hex and 0–255 values are normalized to 0–1 by `/255`. Malformed or empty input preserves the previous defaults (`0.0` for RGB, `1.0` for alpha), so nothing can regress. It is a free function rather than a `static` method on the `Decodable` `ColorComponents` struct, because adding a static member breaks Skip's reflection-based JSON decoding of `ColorSet`.

The hex prefix is detected **explicitly** rather than relying on `Double(_:)` to reject hex, because Swift parses `Double("0xF1")` as `241.0` while the transpiled Kotlin `"0xF1".toDoubleOrNull()` is `null`. A "try `Double` first" approach would therefore behave differently across platforms (and is precisely why this bug is Android-only); prefix detection keeps Swift and Kotlin identical.

The **8-bit (0–255)** decimal method (e.g. `"148"`) is distinguished from floating point by the absence of a decimal point — Xcode writes floats as `"1.000"` and 8-bit as bare integers — so it's handled too, since the issue asks to cover all Input Method variants. (Grayscale colorsets, which use `white`/`alpha` components instead of RGB, are a separate pre-existing gap not addressed here.)

## Before / After evidence

`parseColorComponent` was exercised over 13 cases spanning all three Input Methods (floating point, `0x`/`0X`/`#` hex, lowercase hex, 8-bit `148`/`255`/`4`, empty/nil, malformed):

```
PASS | float passthrough: 0.945
PASS | hex 0x lowercase: 0.9451    (before: toDoubleOrNull -> null -> 0.0 -> black on Android)
PASS | hex # prefix: 0.5333
PASS | 8-bit (0-255): 0.5804       (148/255; before: 148.0 -> clamped to white)
PASS | 8-bit (0-255) max: 1.0
PASS | malformed -> default: 0.0
...
13/13 cases passed
hex 0x04/0xF1/0x88, 8-bit 4/241/136, and float 0.0157/0.9451/0.5333 all resolve to the same color
```

`swift test --filter ColorTests` → **2 tests, 0 failures** (native build clean; existing tests unaffected — the template's required check). `skip test --filter ColorTests` (transpiled Kotlin + Robolectric) → **succeeded**, confirming the fix compiles/transpiles correctly and the suite stays green. The three new render tests (`DISABLEDtestHexColorset` / `DISABLEDtestFloatColorset` / `DISABLEDtestIntColorset`) assert all three encodings render `#04F188`; they are `DISABLED`-prefixed because decoding a bundled component `.colorset` throws a kotlin-reflect error under the Robolectric *unit* runner (it works on a real device) — run them on an emulator/device.

## Acceptance criteria
- [x] Tests added that target the fix (`DISABLEDtestHexColorset`/`DISABLEDtestFloatColorset`/`DISABLEDtestIntColorset`, one per Input Method); the parsing algorithm is also covered by a standalone 13-case logic check
- [x] Existing test suite passes (`swift test` 2/2; `skip test` on Robolectric succeeded)
- [x] Follows project style (matches surrounding `Color.swift` / `ColorTests` conventions)
- [x] No breaking changes (float path unchanged; defaults preserved)

Closes #146

-----

Skip Pull Request Checklist:

- [x] REQUIRED: I have checked whether this change requires a corresponding update in Skip Fuse UI — none needed (internal colorset parsing; no public API change)
- [ ] OPTIONAL: I have added an example of any UI changes to the Showcase sample app

-----

- [x] AI was used to assist with generating this PR. **How:** I used Claude Code as a pair-programmer to locate the root cause in `ColorComponents`, draft the `parseColorComponent` helper, and scaffold the fixtures and tests. **Manual verification:** I read and understood every line, empirically confirmed the `Double("0xF1")` Swift-vs-Kotlin parsing difference that drove the prefix-first design, traced the `#if SKIP` code path, and ran `swift test` and `skip test` (Robolectric) plus a standalone 10-case logic check to confirm the before/after behavior. The local `skip test` run also surfaced two transpilation issues I then fixed: Swift's `Int(_:radix:)` does not transpile (switched to Kotlin's `toIntOrNull(radix:)`), and a `static` helper on the `Decodable` struct broke `ColorSet` decoding (moved it to a free function).

### DIFF
--- a/Sources/SkipUI/SkipUI/Color/Color.swift
+++ b/Sources/SkipUI/SkipUI/Color/Color.swift
@@ -623,13 +623,44 @@ private struct ColorSet : Decodable {
         let alpha: String?
 
         var color: Color {
-            let redValue = Double(red ?? "") ?? 0.0
-            let greenValue = Double(green ?? "") ?? 0.0
-            let blueValue = Double(blue ?? "") ?? 0.0
-            let alphaValue = Double(alpha ?? "") ?? 1.0
+            let redValue = parseColorComponent(red, defaultValue: 0.0)
+            let greenValue = parseColorComponent(green, defaultValue: 0.0)
+            let blueValue = parseColorComponent(blue, defaultValue: 0.0)
+            let alphaValue = parseColorComponent(alpha, defaultValue: 1.0)
             return Color(red: redValue, green: greenValue, blue: blueValue, opacity: alphaValue)
         }
     }
+}
+
+/// Parse a single `.colorset` channel, covering all of Xcode's numeric "Input Method" variants:
+/// Floating Point ("0.945", passed through unchanged), 8-bit Hexadecimal ("0xF1" / "#F1"), and
+/// 8-bit (0-255) ("241"). Hex bytes and 0-255 integers are normalized to 0...1 by dividing by 255.
+/// Malformed or empty input falls back to `defaultValue`. See https://github.com/skiptools/skip-ui/issues/146
+///
+/// The hex prefix is detected explicitly rather than relying on `Double(_:)` to reject it: Swift
+/// parses `Double("0xF1")` as `241.0`, whereas the transpiled Kotlin `"0xF1".toDoubleOrNull()` is
+/// `null`. A "try Double first" approach would therefore behave differently once transpiled (and is
+/// the reason #146 manifests only on Android). Checking the prefix keeps both platforms identical.
+/// Hex digits are parsed with Kotlin's `toIntOrNull(radix:)` because Swift's `Int(_:radix:)` does not
+/// transpile. This is a free function rather than a method on `ColorComponents` because adding a
+/// `static` member to that `Decodable` type breaks Skip's reflection-based JSON decoding of `ColorSet`.
+private func parseColorComponent(_ string: String?, defaultValue: Double) -> Double {
+    guard let string, !string.isEmpty else { return defaultValue }
+    if string.hasPrefix("0x") || string.hasPrefix("0X") {
+        let byte: Int? = String(string.dropFirst(2)).toIntOrNull(radix: 16)
+        if let byte { return Double(byte) / 255.0 }
+        return defaultValue
+    }
+    if string.hasPrefix("#") {
+        let byte: Int? = String(string.dropFirst(1)).toIntOrNull(radix: 16)
+        if let byte { return Double(byte) / 255.0 }
+        return defaultValue
+    }
+    // Floating Point emits a decimal ("0.580"); the "8-bit (0-255)" method emits a bare integer ("148").
+    if !string.contains("."), let intValue = Int(string) {
+        return Double(intValue) / 255.0
+    }
+    return Double(string) ?? defaultValue
 }
 
 #endif
--- a/Tests/SkipUITests/ColorTests.swift
+++ b/Tests/SkipUITests/ColorTests.swift
@@ -14,6 +14,42 @@ final class ColorTests: XCSnapshotTestCase {
         XCTAssertEqual("F", try render(compact: 1, view: Color.white.frame(width: 1.0, height: 1.0)).pixmap)
     }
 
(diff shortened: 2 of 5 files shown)


arulagarwal:
@cla-bot check

marcprux:
I think this is obviated by https://github.com/skiptools/skip-ui/pull/471, but feel free to take a look and see if there are any other improvements from the PR that might be applicable (like the tests).

Issue #41, opened 2024-08-18
Unsupported ScrollView modifiers for view aligned / paging behavior...


jeffc-dev:
Skip does not currently support the following modifiers to ScrollView and/or its contents:

`scrollTargetBehavior` (`.viewAligned`, `.paged`, etc) [Link](https://developer.apple.com/documentation/swiftui/view/scrolltargetbehavior(_:))
`scrollClipDisabled` [Link](https://developer.apple.com/documentation/swiftui/view/scrollclipdisabled(_:))
`scrollTargetLayout` [Link](https://developer.apple.com/documentation/swiftui/view/scrolltargetbehavior(_:))
`.scrollPosition(id:anchor:)` [Link](https://developer.apple.com/documentation/swiftui/view/scrollposition(id:anchor:))

Without these, it is not easily possible to create _paged_ or _view-aligned_ scroll views (esp. with subviews that "peek" through on either side), or with a position that can be controlled programmatically.

Common examples of this sort of behavior in use can be found in the App Store, TV, Books, and Music apps. I've attached an example from the TV app as reference.

![image](https://github.com/user-attachments/assets/520ce2c0-d303-4a40-b485-dbbcef12e811)

I don't know if this is easily doable with Jetpack Compose, but if feasible it'd be nice to have it supported.

dfabulich:
In Jetpack Compose, you can do something similar by setting the `flingBehavior` of `LazyRow`.

dfabulich:
I’m prepared to write some Kotlin to make this work, but it's not clear what the optimal way to go about it is. I'd appreciate some feedback.

I can see a few different options here:

1. Maybe I'd try to cook a separate dual-platform framework for scroll snapping, designed to follow each platform's conventions.

    On iOS, you'd pass it a collection of views and an alignment (horizontal or vertical) and it would return a ScrollView with `.scrollTargetBehavior(.viewAligned)` containing an Lastack with `.scrollTargetLayout()`.

    On Android, it would return a `LazyRow` or `LazyColumn` with its `FlingBehavior` set to snap.

1. I could try to learn how to hack on SkipUI and implement SwiftUI's `scrollTargetBehavior` in Kotlin. This sounds pretty hard, especially since I'm not very experienced working with Skip, but also because the shape of the APIs are quite different.

    SwiftUI lets you specify which views in a ScrollView are (or are not) scroll targets.

[...]

PR #12, opened 2023-11-20
Display toolbar for toolbar items at root

pnewell:
Fixes bug where top navigation bar is not displayed when in a root view without a navigationTitle but with leading/trailing toolbar items.

Issue #10, opened 2023-11-10
Can't support .onAppear, .animation, .overlay ?


rajaslucky89:
why **skip** not support `.onAppear {}`, `.animation(.easeInOut(duration: 1.3), value: splashScreen)` and `.overlay` ?

aabewhite:
We simply haven't gotten to these modifiers yet! Note that you can use .task to a lot of the things .onAppear can do. And you can use unsupported modifiers for your iOS app with #if !SKIP. They just won't appear in your Android app.

`SomeView()
    #if !SKIP
    .animation(...)
    #endif
`

aabewhite:
.onAppear is now supported

aabewhite:
.overlay is also supported, and .animation is on our radar

PR #475, opened 2026-07-03
Bugfix: Stabilize NavigationStack Layout

fhasse95:
This PR stabilizes `NavigationStack` layout behavior so that the navigation bar and the content below it no longer jump when the navigation bar height is measured or updated.

Before this change, the navigation bar could be laid out with an initial height of zero or an incomplete height, and then expand to its final size after the first layout/measurement pass. Visually, this made the navigation bar appear collapsed immediately after a tab switch and then "slide" or jump into place once layout completed.

In release builds with simple views this is often hard to notice because the layout pass completes very quickly. With more complex views, or in debug builds, however, the intermediate state becomes visible. This is also reproducible in the Skip Showcase app in a debug build when taking a closer look.

To reproduce this in the Skip Showcase app:
1. Install and launch a debug build of the app
2. Start on the `About` tab
3. Switch to the `Showcase` tab
4. Observe the navigation bar during the initial layout of the `Showcase` view

Note: The jump only happens during the first layout of that view. After the `Showcase` tab has been laid out once, switching away and back again usually does not reproduce the issue since the navigation bar has already been measured.

Below is a before/after comparison of this example in the Skip Showcase app, slowed down to make the issue easier to see:

**Before:**
Before

**After:**
After

As shown above, before the change, the navigation bar is initially collapsed after switching tabs and becomes visible only after layout catches up. After the change, the navigation bar is present at the correct size immediately.

This is achieved by reserving an estimated top-bar height before the first real measurement is available. Once the actual top-bar size has been measured, the layout switches to the measured value. This keeps the content and navigation bar stable during initial layout and prevents the visible jump.

I also manually verified additional cases that previously showed navigation-bar jumps during initial layout, including `NavigationStack` inside `.sheet` and `.fullScreenCover`, using my own Skip app with more complex views and navigation flows. Those cases are also fixed with this change.

As part of the same layout improvements, this PR also reduces unnecessary empty space between the navigation bar and the top edge of sheets, making better use of the available sheet area:

**Large Title (Before / After):**
Before (Sheet Large)After (Sheet Large)

**Inline Title (Before / After):**
Before (Sheet Inline)After (Sheet Inline)

---

Thank you for contributing to the Skip project! Please review the contribution guide at https://skip.dev/docs/contributing/ for advice and guidance on making high-quality PRs.

Use this space to describe your change and add any labels (bug, enhancement, documentation, etc.) to help categorize your contribution.

Skip Pull Request Checklist:

- [X] REQUIRED: I have checked whether this change requires a corresponding update in the [Skip Fuse UI](https://github.com/skiptools/skip-fuse-ui) repository (link related PR if applicable)

-----

- [X] AI was used to generate or assist with generating this PR. *Please specify below how you used AI to help you, and what steps you have taken to manually verify the changes*.

I used Codex to iteratively investigate and refine the fix for the NavigationStack layout issue. To verify the results, I tested multiple playground views in the Skip Showcase app and my own Skip app, which contains several navigation scenarios, including sheets, full-screen covers, tab bars, and nested navigation. All of these scenarios work properly after the bug fix.

-----

### DIFF
--- a/Sources/SkipUI/SkipUI/Compose/ComposeLayouts.swift
+++ b/Sources/SkipUI/SkipUI/Compose/ComposeLayouts.swift
@@ -158,8 +158,9 @@ private func flexibleLayoutFloat(_ value: CGFloat?) -> Float? {
 ///     the given closure as a pixel rect.
 /// - Parameter checkEdges: Which edges to check to see if we're against a safe area. Any matching edges will be
 ///     passed to the given closure.
+/// - Parameter bottomAdjacencyTolerancePx: Distance from the bottom safe boundary still considered adjacent.
 /// - Parameter logTag: When non-empty, emits Android ``Log`` lines with tag `SkipUI.ISAL.<logTag>` (e.g. filter logcat `SkipUI.ISAL.List`).
-@Composable func IgnoresSafeAreaLayout(expandInto: Edge.Set, checkEdges: Edge.Set = [], modifier: Modifier = Modifier, logTag: String = "", target: @Composable (IntRect, Edge.Set) -> Void) {
+@Composable func IgnoresSafeAreaLayout(expandInto: Edge.Set, checkEdges: Edge.Set = [], modifier: Modifier = Modifier, bottomAdjacencyTolerancePx: Float = Float(0.1), logTag: String = "", target: @Composable (IntRect, Edge.Set) -> Void) {
     guard let safeArea = EnvironmentValues.shared._safeArea else {
         if !logTag.isEmpty {
             Log.d("SkipUI.ISAL.\(logTag)", "no SafeArea in environment; skipping expansion")
@@ -220,7 +221,7 @@ private func flexibleLayoutFloat(_ value: CGFloat?) -> Float? {
     } in: {
         Layout(modifier: modifier.onGloballyPositionedInWindow {
             let probeEdges = expandInto.union(checkEdges)
-            let newEdges = adjacentSafeAreaEdges(bounds: $0, safeArea: safeArea, isRTL: isRTL, checkEdges: probeEdges)
+            let newEdges = adjacentSafeAreaEdges(bounds: $0, safeArea: safeArea, isRTL: isRTL, checkEdges: probeEdges, bottomTolerancePx: bottomAdjacencyTolerancePx)
             if !logTag.isEmpty {
                 let previous = edgesState.value
                 if newEdges != previous {
@@ -255,12 +256,12 @@ private func flexibleLayoutFloat(_ value: CGFloat?) -> Float? {
     }
 }
 
-private func adjacentSafeAreaEdges(bounds: Rect, safeArea: SafeArea, isRTL: Bool, checkEdges: Edge.Set) -> Edge.Set {
+private func adjacentSafeAreaEdges(bounds: Rect, safeArea: SafeArea, isRTL: Bool, checkEdges: Edge.Set, bottomTolerancePx: Float = Float(0.1)) -> Edge.Set {
     var edges: Edge.Set = []
     if checkEdges.contains(Edge.Set.top), bounds.top <= safeArea.safeBoundsPx.top + 0.1 {
         edges.insert(Edge.Set.top)
     }
-    if checkEdges.contains(Edge.Set.bottom), bounds.bottom >= safeArea.safeBoundsPx.bottom - 0.1 {
+    if checkEdges.contains(Edge.Set.bottom), bounds.bottom >= safeArea.safeBoundsPx.bottom - bottomTolerancePx {
         edges.insert(Edge.Set.bottom)
     }
     if isRTL {
--- a/Sources/SkipUI/SkipUI/Compose/ComposeLayouts.swift
+++ b/Sources/SkipUI/SkipUI/Compose/ComposeLayouts.swift
@@ -169,6 +169,11 @@ private func flexibleLayoutFloat(_ value: CGFloat?) -> Float? {
         return
     }
 
+    let effectiveBottomAdjacencyTolerancePx = max(
+        bottomAdjacencyTolerancePx,
+        safeArea.presentationBoundsPx.bottom - safeArea.safeBoundsPx.bottom
+    )
+
     if !logTag.isEmpty {
         LaunchedEffect(logTag, expandInto.rawValue, checkEdges.rawValue) {
             Log.d("SkipUI.ISAL.\(logTag)", "init expandInto=\(expandInto) checkEdges=\(checkEdges) edgesState(initial)=\(checkEdges)")
@@ -221,7 +226,7 @@ private func flexibleLayoutFloat(_ value: CGFloat?) -> Float? {
     } in: {
         Layout(modifier: modifier.onGloballyPositionedInWindow {
             let probeEdges = expandInto.union(checkEdges)
-            let newEdges = adjacentSafeAreaEdges(bounds: $0, safeArea: safeArea, isRTL: isRTL, checkEdges: probeEdges, bottomTolerancePx: bottomAdjacencyTolerancePx)
+            let newEdges = adjacentSafeAreaEdges(bounds: $0, safeArea: safeArea, isRTL: isRTL, checkEdges: probeEdges, bottomTolerancePx: effectiveBottomAdjacencyTolerancePx)
             if !logTag.isEmpty {
                 let previous = edgesState.value
                 if newEdges != previous {
--- a/Sources/SkipUI/SkipUI/Containers/List.swift
+++ b/Sources/SkipUI/SkipUI/Containers/List.swift
@@ -156,8 +156,9 @@ public final class List : View, Renderable {
         let safeArea = EnvironmentValues.shared._safeArea
         var ignoresSafeAreaEdges: Edge.Set = [.top, .bottom]
         ignoresSafeAreaEdges.formIntersection(safeArea?.absoluteSystemBarEdges ?? [])
+        let bottomAdjacencyTolerancePx = safeArea.map { max(Float(0.1), $0.presentationBoundsPx.bottom - $0.safeBoundsPx.bottom) } ?? Float(0.1)
         ComposeContainer(scrollAxes: .vertical, modifier: context.modifier, fillWidth: true, fillHeight: true, then: Modifier.background(BackgroundColor(styling: styling, isItem: false))) { modifier in
-            IgnoresSafeAreaLayout(expandInto: ignoresSafeAreaEdges, checkEdges: [.bottom], modifier: modifier, logTag: "List") { safeAreaExpansion, safeAreaEdges in
+            IgnoresSafeAreaLayout(expandInto: ignoresSafeAreaEdges, checkEdges: [.bottom], modifier: modifier, bottomAdjacencyTolerancePx: bottomAdjacencyTolerancePx, logTag: "List") { safeAreaExpansion, safeAreaEdges in
                 var containerModifier: Modifier
                 let refreshing = remember { mutableStateOf(false) }
                 let refreshAction = EnvironmentValues.shared.refresh
--- a/Sources/SkipUI/SkipUI/Containers/List.swift
+++ b/Sources/SkipUI/SkipUI/Containers/List.swift
@@ -156,9 +156,8 @@ public final class List : View, Renderable {
         let safeArea = EnvironmentValues.shared._safeArea
         var ignoresSafeAreaEdges: Edge.Set = [.top, .bottom]
         ignoresSafeAreaEdges.formIntersection(safeArea?.absoluteSystemBarEdges ?? [])
-        let bottomAdjacencyTolerancePx = safeArea.map { max(Float(0.1), $0.presentationBoundsPx.bottom - $0.safeBoundsPx.bottom) } ?? Float(0.1)
         ComposeContainer(scrollAxes: .vertical, modifier: context.modifier, fillWidth: true, fillHeight: true, then: Modifier.background(BackgroundColor(styling: styling, isItem: false))) { modifier in
-            IgnoresSafeAreaLayout(expandInto: ignoresSafeAreaEdges, checkEdges: [.bottom], modifier: modifier, bottomAdjacencyTolerancePx: bottomAdjacencyTolerancePx, logTag: "List") { safeAreaExpansion, safeAreaEdges in
+            IgnoresSafeAreaLayout(expandInto: ignoresSafeAreaEdges, checkEdges: [.bottom], modifier: modifier, logTag: "List") { safeAreaExpansion, safeAreaEdges in
                 var containerModifier: Modifier
                 let refreshing = remember { mutableStateOf(false) }
                 let refreshAction = EnvironmentValues.shared.refresh
--- a/Sources/SkipUI/SkipUI/Containers/Navigation.swift
+++ b/Sources/SkipUI/SkipUI/Containers/Navigation.swift
@@ -76,6 +76,7 @@ import androidx.compose.ui.geometry.Rect
 import androidx.compose.ui.graphics.graphicsLayer
 import androidx.compose.ui.input.nestedscroll.nestedScroll
 import androidx.compose.ui.layout.boundsInWindow
+import androidx.compose.ui.layout.onGloballyPositioned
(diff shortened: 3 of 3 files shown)


dfabulich:
This code is _extremely_ finicky, and has tests all over the showcase. I think it would help for you to list out exactly which playgrounds you tested and how.

One that you didn't explicitly call out is the SafeArea playground --> Geometry padding playground, which has facilities for hiding and showing the Navigation Bar, both with and without a sheet.

fhasse95:
> This code is _extremely_ finicky, and has tests all over the showcase. I think it would help for you to list out exactly which playgrounds you tested and how.
> 
> One that you didn't explicitly call out is the SafeArea playground --> Geometry padding playground, which has facilities for hiding and showing the Navigation Bar, both with and without a sheet.

Yes, I totally agree. Since this code is central and used in many places, I wanted to be extra careful. That's why I tried out the changes in various scenarios.

In Skip Showcase, I validated the changes using the following playgrounds:

* **Sheet Playground**
  * Tested all options, including both `.sheet` and `.fullScreenCover`, with both `inline` and `large` navigation bar title modes.

* **NavigationStack Playground**
  * Tested all available options, again with both `inline` and `large` navigation bar title modes.

* **Toolbar Playground**
  * Tested all available options, including hiding/showing the navigation bar, custom navigation bar colors, etc.

* **SafeArea Playground**
  * Tested all available options, with particular focus on `.fullScreenCover`, `.sheet`, and the Geometry Padding playground.
  * I also verified that features such as hiding/showing the navigation bar continued to behave correctly there.

During implementation, I first tested the changes in my own app across a variety of scenarios, refining the implementation iteratively based on my observations. Besides fixing the original issue, I specifically looked at things like switching between tabs, verifying the initial layout pass of the navigation bar, and the layout behavior of sheets and full-screen covers. I also made sure that neither the navigation bar nor the underlying content jumped during transitions.

Most of my testing was done on my Google Pixel 9, where I also verified gesture-driven interactions, such as slowly dismissing a sheet using the back gesture, to ensure the navigation bar behaves correctly throughout the interactive transition.

Once everything looked good there, I built Skip Showcase against my local SkipUI branch and went through all of the playgrounds above. To make sure I hadn't introduced any regressions, I repeatedly switched between my bugfix branch and the current `main` branch of SkipUI and compared the behavior side by side. Other than the initial layout pass bug now being fixed, I wasn't able to observe any side effects.

That said, I obviously can't completely rule out that I may have missed an edge case. So please feel free to try it with either your own app(s) or with Skip Showcase, and let me know if you spot anything that doesn't look right 😊

fhasse95:
@dfabulich I spent some more time testing this change extensively and found one additional edge case, which is now fixed by the latest commit.

The issue occurred when the actual toolbar content was taller than the initially reserved navigation bar height. In that situation, the sheet could still jump slightly during the interactive Android back gesture while the sheet or full-screen cover was shrinking.

I encountered this in my own app while testing different localizations. For example, the Arabic version of my Save button resulted in slightly taller toolbar content than the default content used in some other languages:




This caused the measured toolbar height to exceed the initial estimate, making the transition visible during the gesture. The latest commit accounts for this case, and I can no longer reproduce the jump.

**To further clarify the main issue addressed by this PR:**

Before this change, the navigation bar was initially laid out with a height of zero and would then expand to its final measured size during a later layout pass. With smaller or simpler views, this is often difficult to notice because layout completes very quickly. In larger and more complex applications, particularly in debug builds, however, the intermediate layout state becomes clearly visible.

This PR avoids that intermediate state by reserving an estimated navigation bar height until the actual toolbar height has been been measured. Once the measurement is available, the estimate is seamlessly replaced with the final value.

This behavior can be observed whenever a view containing a `NavigationStack` is presented for the first time.

For example:

1. After switching tabs (and even right after app launch), the navigation bar may initially not be rendered at all. It then appears from the top once its height has been measured, causing the content below it to jump.

2. When presenting a `.sheet` or `.fullScreenCover`, the navigation bar may again initially be laid out with a height of zero. During that brief state, the content shifts and the navigation bar background is missing, exposing a temporary white gap underneath the top bar.

Both behaviors can also be reproduced in the Skip Showcase app. Since it is a relatively simple application, the layout issues are only briefly visible and are easiest to spot by recording the screen and stepping through the captured frames. In more complex applications, however, they become much more obvious. The tab-switching issue is visible when opening a tab for the first time, and the presentation issue can be observed in the Sheet Playground.

I have attached relevant frame-by-frame screenshots from Skip Showcase below:

**Switching Tabs [Before]**





**Switching Tabs [After]**




> As you can see, the navigation bar is already laid out with the correct height before the transition has completed.

**Presenting Sheets [Before]**





> The same issue occurs when presenting sheets. Initially, the navigation bar is laid out with a height of zero before receiving its final height during a later layout pass.

**Presenting Sheets [After]**




> After the changes, the navigation bar already has the correct height even before the sheet presentation has completed.

**Presenting Fullscreen Covers [Before]**





**Presenting Fullscreen Covers [After]**




> With full-screen covers, the underlying issue was exactly the same. In addition, the navigation bar background was temporarily missing during the transition, resulting in a visible white gap. This PR fixes both issues.

Below are also a few screen recordings from my own Skip app running in debug mode. Since the UI is more complex (icons, gradients, additional drawing work, etc.), the delayed layout becomes much more noticeable than in the Skip Showcase app, as mentioned earlier.

**Switching Tabs [Before]**



**Switching Tabs [After]**



**Presenting Fullscreen Covers [Before]**



**Presenting Fullscreen Covers [After]**



I hope the additional screenshots and explanations help clarify the issue. If you have any further questions or concerns, please let me know 😊

marcprux (approved):
This looks good to me on the surface, but I do share @dfabulich's concerns over the (lamentable) intricacy of navigation on Android, so I'd like to get his 2nd opinion before I merge.

fhasse95:
@marcprux @dfabulich FYI: I merged this branch with the latest `main` again and did another round of testing.

While testing, I noticed two additional issues. First, my previous version still contained a bug that caused the bottom safe area inset to be applied incorrectly for sheets. Second, I found another layout issue in **SafeAreaPlayground → Sheet → Bottom Toolbar** during the interactive Android back dismiss gesture, which is also present on the current `main` branch:

**Before:**
Before

**After:**
After

> As you can see, before the fix the safe area was recalculated with a slight delay during the back dismiss gesture, causing the bottom toolbar to jump slightly.

Both issues are fixed in the latest commit.

dfabulich:
I've put some time and thought into this PR and, paradoxically, I agree that it fixes a bug, but I'm not sure we should go in this direction.

There are two core problems with SkipUI's `NavigationStack`.

1. **Problem 1: `NavigationStack` preferences vs. `TopAppBar` parameters:** One core problem is that Compose [`TopAppBar`](https://developer.android.com/develop/ui/compose/components/app-bars) requires all of its configuration to be passed in as it's composed, in parameters and modifiers, but SwiftUI `NavigationStack` expects all of its configuration to be set via preference modifiers on the body content. SkipUI renders the `NavigationStack`, typically painting something "wrong", then, as the body composes, discovers what it should have done, recomposes, and repaints.

    My attempt to address Problem 1 was to add [layout hints](https://skip.dev/docs/modules/skip-ui/#prevent-layout-shifting-with-hints). That sucks, because it means users have to add some special Skip modifiers to prevent a layout shift. Ideally, SkipUI `NavigationStack` would "just work" without hints.
    
    @fhasse95 have you tried layout hints? (You made no mention of hints in this PR.)

2. **Problem 2: The body's layout depends on the layout of the `TopAppBar`, which we can't reliably predict:** Even if we pass in hints, so we can render the `TopAppBar` armed with the knowledge of the title and the title display mode, we still have no way of knowing for sure how tall the `TopAppBar` will be. Will it be 62dp collapsed? 112dp? Well, has the user selected an enlarged accessibility font? Has the user chosen an OS override font? We can make some guesses, but the only way to be sure is to actually render the `TopAppBar` and _measure_ its layout.

    Today, we handle Problem 2 with an `onGloballyPositioned` modifier, but that guarantees that we'll get one wrong paint before we get a chance to recompose and render the body in the correct layout, even with layout hints.
    
The core idea of the initial commit of this PR #475 is to make _better estimates_ about how tall the `TopAppBar` will be, addressing Problem 2 in a different way.

I have a half-baked PR on my machine that did something similar, but I wound up not filing it, partly because Problem 2 is invisible on my test devices as long as I pass in layout hints. (It's still _happening_ … it's still adding to _jank_. But it's not _visible_.)

@fhasse95's estimates are pretty decent; when they're wrong, they're less wrong than the current behavior.

But adding better estimates means adding _more code_ to `Navigation.swift`, making it even _more_ complex. And it still has [no automated tests](https://github.com/skiptools/skip-ui/issues/492), which IMO is a big part of the reason we kept you waiting almost a month to get this thing reviewed and merged.

I reviewed this PR and thought, "This _can't_ be the right way to handle this. More and more and more code… this can't be right." `Navigation.swift` already had a "code smell" before you started to work on it; this PR makes that smell more visible. (More… pungent??)

So, I went and looked at how Compose [`Scaffold`](https://developer.android.com/develop/ui/compose/components/scaffold) handles this. The answer? [`SubcomposeLayout`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/Scaffold.kt;l=163?q=scaffold&ss=androidx%2Fplatform%2Fframeworks%2Fsupport:compose%2F). Subcomposing allows `Scaffold` to measure the `TopAppBar` and use those measurements to position the body content in a single pass. It doesn't attempt to _estimate_ how large the `TopAppBar` will be, because it doesn't need to.

So, I pulled together a draft PR in that direction.
* #494

There are other fixes in the current PR #475, some of which (I think?) don't have anything directly to do with the new estimation code; it's hard to evaluate those in isolation. (I know from experience that fixes to `Navigation.swift` inherently pile up on one another, so it makes sense that you've added more fixes in here as you found them. Surely someday Github will give us all access to [stacked PRs](https://github.github.com/gh-stack/).)

So, we have a couple options:

1. Merge this PR as-is. It'll make Navigation.swift more complex, and the estimates can't be perfect
2. Switch to a `SubcomposeLayout` approach like #494. To be honest, #494 doesn't _look_ any simpler, but that's because I put it behind a `layoutImplementationVersion` flag, allowing users to switch back to older implementations at runtime. Maybe someday we'd release a SkipUI 2.0, remove that flag, and simplify.

WDYT?

fhasse95:
@dfabulich Thanks for taking the time to review this so thoroughly and for explaining the underlying issues in such detail. I also appreciate you putting together #494.

To answer your question: I haven't tried layout hints yet, so I can't really compare the behavior with and without them. However, if the `SubcomposeLayout` approach doesn't end up being significantly simpler overall, my preference would be to simply merge this PR as it is and then revisit the Navigation implementation as part of a future SkipUI 2.0 refactor, similar to what you suggested with the `layoutImplementationVersion` flag.

That way, we'd at least have the current bugs fixed in the existing implementation, even if `Navigation.swift` remains fairly complex, which, as you pointed out, was already the case before this PR.

If that sounds reasonable to you, I'd merge the latest `main` branch into this PR first, since another contributor has already created a fix for the sheet top spacing issue that is also included in this PR. After that, however, I think this PR would be ready to merge.

dfabulich:
A few thoughts.

1. I think it would be best to rebase your branch onto `main` and retest, rather than just merging `main` into your branch and resolving conflicts.
1. I think @marcprux has to be the decider for architectural direction, both for this PR and my `SubcomposeLayout` PR, and, perhaps most importantly, for #492.
1. I don't think anyone can simply read the code of this PR or my `SubcomposeLayout` PR and say "oh, yeah, _this_ approach is correct." Making this decision will require testing. Since we don't have good automated tests, it will require manual testing. So, I think it is worth your time to rebase this PR onto `main`, if only so we can manually test it and compare it against the other PR. (I might also suggest doing this with a [stacked PR](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests), which is available now to all users, at long last.)
1. I think `SubcomposeLayout` will be simpler than the current implementation, but only when we can kill old `layoutImplementationVersion` implementations. If we _don't_ kill old implementations, we can only ever monotonically increase complexity, which seems bad.
1. IMO `layoutImplementationVersion` exists because we don't have adequate tests. If we had good tests, we could refactor implementations with the confidence of knowing that all of the tests still pass.

fhasse95:
Yes, that makes sense to me.

I also took some time to briefly test the current state of your subcompose layout / safe area refactoring branch with both my app and the Skip Showcase app so I could compare the two approaches.

From what I observed, it already improves a few scenarios. For example, `fullScreenCover` and `sheet` no longer cause the content to slide underneath the navigation bar, which were also among the issues addressed by this PR.

However, I can still reproduce some of the layout issues that I originally described, such as the initial navigation bar/top bar slide-in during the first layout pass. I also noticed a few additional layout inconsistencies during tab switching:

**Tab Switching / Initial Layout Pass [[Main Branch](https://github.com/skiptools/skip-ui)]**


**Tab Switching / Initial Layout Pass [[Safe Area Refactoring Branch](https://github.com/dfabulich/skip-ui/tree/navigation-safe-area-refactor)]**


**Tab Switching / Initial Layout Pass [[Navigation Layout Branch (This PR)](https://github.com/fhasse95/skip-ui/tree/Navigation-Layout-Bugfix)]**


In contrast, after updating this PR to the latest `main` and retesting it with both my app and the Skip Showcase app, I couldn't reproduce any of the layout issues described in #475 anymore.

Of course, my implementation is still based on estimates, so I can't completely rule out that there are edge cases I haven't considered yet. I also agree that, in the long run, a solid automated testing strategy would make it much easier to validate and confidently refactor the navigation implementation.

That said, based on the scenarios I tested manually, this PR already provides meaningful improvements for existing SkipUI users by resolving the layout issues within the current navigation implementation, while a more comprehensive refactoring and automated test strategy can still be developed over time.

What do you think, @marcprux? I'd be interested to hear your thoughts and where you think we should go from here.

marcprux:
> What do you think, @marcprux? I'd be interested to hear your thoughts and where you think we should go from here.

I'm on vacation now, but I'll be getting back later this week and caching up on the backlog, including this.

For now, my only thought is that I appreciate the thoughtful conversation around this issue thus far, as well as the explanations (and screenshots 😍). I'm certainly in favor of improvements like this. At the same time, we're growing more conservative about PRs this large due to the potential to break existing apps. But I'll have more to say soon!

fhasse95:
Thanks, Marc! I really appreciate you taking the time to look into this. Enjoy the rest of your vacation, and I look forward to hearing your thoughts once you're back 😊

fhasse95:
Hi @marcprux, just wanted to follow up on this one now that some time has passed. Have you had a chance to take another look at it, or is there anything you'd like me to adjust to make the PR easier to review or merge?

I also still have a few other PRs open in skip-ui ([#507](https://github.com/skiptools/skip-ui/pull/507), [#513](https://github.com/skiptools/skip-ui/pull/513), [#519](https://github.com/skiptools/skip-ui/pull/519)), as well as across some of the other Skip repositories, such as skip-kit ([#41](https://github.com/skiptools/skip-kit/pull/41)), skip-firebase ([#101](https://github.com/skiptools/skip-firebase/pull/101)), and skip-revenue ([#10](https://github.com/skiptools/skip-revenue/pull/10), [#11](https://github.com/skiptools/skip-revenue/pull/11), [#12](https://github.com/skiptools/skip-revenue/pull/12)). No rush on those, but I just wanted to make sure nothing has fallen off the radar 🙂

ricardssx:
Hi all,
Just wanted to add the impact on our side-

We’re seeing this on a ZTE A71 5G in a release build. `navigationStackLayoutHints` gets the title in on the first frame, but the content still shifts once the top bar is measured. I tested 2fd17cb in our emulator build and the tab/menu transitions looked fixed.

We’re considering pinning this commit as it’s semi-severe for prod.

marcprux:
This is now available in `skip-iu` [1.59.3](https://github.com/skiptools/skip-ui/releases/tag/1.59.3).

Sorry for the very long delay in getting this merged! As well as having a huge backlog, I'm also aligned with concerned about the increase in complexity. But it does seem to work fairly well to solve the issue, and the translation of navigation concepts from SwiftUI to Compose is inherently a very complex ball of glue. Thanks for the contribution and the persistence!

Issue #492, opened 2026-07-25
NavigationStack and TabView have inadequate automated testing


dfabulich:
NavigationStack and TabView are _incredibly_ complicated pieces of code; it's extremely difficult to touch them without accidentally breaking something.

skip-ui has some automated tests, but no automated tests of these areas. We should fix that, ideally not (merely) by adding automated tests to https://github.com/skiptools/skipapp-showcase but adding tests that run during the skip-ui CI.

dfabulich:
There are a _lot_ of open questions about how this might be best accomplished.

* Would these be tests that run on iOS and Android?
  * If so, what cross-platform testing library would we use?
    * Marc mentioned preferring something higher-level like [Maestro](https://maestro.dev/).
    * I think I'd prefer XCUIAutomation https://github.com/skiptools/skip/issues/649
  * But perhaps we shouldn't. skip-ui runs on Android only, and maybe we'd be better off with Android-only behavior UI tests. There are a lot of ways to do that, too. https://developer.android.com/training/testing/ui-tests/behavior 
    * Compose testing API? https://developer.android.com/develop/ui/compose/testing
    * UI Automator? https://developer.android.com/training/testing/ui-automator
    * Compose Preview Screenshot Testing? https://github.com/skiptools/skip/issues/586
* What device should the tests run on?
  * Robolectric?
  * Or an emulator?
    * Which emulated device? (Multiple devices??)
    * How would the emulated device be set up, managed, and launched?
      * Build-managed devices? https://developer.android.com/studio/test/managed-devices
* Would we want to have different answers for these questions for skip-ui vs end-user apps like skipapp-showcase?

It's not totally clear to me how we'd begin to answer any of these questions.

(Another wrinkle: I think we'd want to test that layout shifts aren't happening, which sounds pretty thorny to me.)

Maybe we could do a bake-off or something? Or maybe @marcprux just picks an approach?

PR #494, opened 2026-07-26
Navigation safe area subcompose refactor

### LINKED ISSUE #493: SkipUI should use Compose `WindowInsets` instead of a `_safeArea` environment value
https://developer.android.com/develop/ui/compose/system/evaluate-rulers

> [WindowInsets](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/WindowInsets) is the standard API in Jetpack Compose for handling areas of the screen that are partially or fully obscured by the system UI. These areas include the status bar, navigation bar, and on-screen keyboard.

Currently, SkipUI manages the safe area using an internal `_safeArea` environment value. This creates a lot of tricky busy work, especially in NavigationStack and TabView, making sure that SkipUI's `_safeArea` lines up with `WindowInsets` provided by the system.

It would be simpler and better to make SkipUI code work with `WindowInsets`, adding/consuming padding as we go.

dfabulich:
* Fixes #493 

**This is currently a draft PR partly because I'm not yet sure there are enough tests, and also because I'm not sure this is the right direction.**
* #492

This PR uses `SubcomposeLayout` to layout the `TopAppBar` and the body content of the `NavigationStack`, mirroring the implementation of Compose [`Scaffold`](https://developer.android.com/develop/ui/compose/components/scaffold) (see their [implementation](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/Scaffold.kt;l=163?q=scaffold&ss=androidx%2Fplatform%2Fframeworks%2Fsupport:compose%2F).)

This eliminates the need to use `onGloballyPositioned` to measure the app bar, eliminating layout shift.

Skip Pull Request Checklist:

- [ ] OPTIONAL: I have tested my change on an Android emulator or device
    (Not required)

-----

Cursor generated the core of this PR. I manually tested it with the Showcase app in the SafeAreaPlayground, but I'm not confident I've done enough manual testing; I don't want to invest a lot of energy into it unless we agree that this is something like the right approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3050,12 +3050,12 @@ SkipUI fully supports SwiftUI's various layout mechanisms, including `HStack`, `
 - Expanding elements such as `Spacer` or `.frame(maxWidth: .infinity)` within nested `HStacks` or `VStacks` may measure differently. Try un-nesting stacks to get more SwiftUI-like layout.
 - Views with `.frame(maxWidth:)` or `.frame(maxHeight:)` set to explicit values larger than the parent's actual size may expand beyond the parent container's bounds on Android. To work around this, use a `GeometryReader` to compute the parent's size and set an explicit `.frame(width:)` or `.frame(height:)` instead of `maxWidth` or `maxHeight`. See [Issue #339](https://github.com/skiptools/skip-ui/issues/339).
 
-Note: if your app was developed under an earlier version of Skip and it relies on nuances of older layout behavior, you can apply the Android-only `.layoutImplementationVersion()` modifier. Set this modifier on a `View` hierarchy to simulate the previous behavior:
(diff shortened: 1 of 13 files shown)


timkoss:
We have a related production reproduction in OurApp.
The app has a custom, app-owned header above several titleless NavigationStacks—one stack per tab. On Android, each stack reserves a large blank region immediately below the custom header. The same gap appears on every tab. iOS does not show this gap.
The custom header is not a navigation bar, so the stack appears to be applying the system top inset again below content that already accounts for the app header. This seems different from the first-frame navigation-bar measurement jump described in PR #475; our issue is a persistent gap, not only an initial layout flash.
We tried app-level toolbar visibility changes and temporarily tested a local SkipUI change, but we reverted the experiment because it was not a safe application-level fix.
Does this PR intend to address a NavigationStack placed below external, app-owned chrome? If so, would it be useful to add a regression case containing:
- an external fixed-height header
- a titleless NavigationStack below it
- navigation-bar visibility changes
- push/pop navigation
- tab switching
- verification that the body begins immediately below the external header
We would prefer to follow the upstream direction rather than maintain a permanent SkipUI fork.
Skip pr image

PR #499, opened 2026-07-29
Guard geometry state writes against sub-pixel bounds jitter to prevent idle recomposition loops

### LINKED ISSUE #488: GeometryReader/onGloballyPositioned bounds writes have no epsilon — sub-pixel Rect jitter causes idle recomposition loops
## Environment

- skip 1.9.4, skip-fuse-ui 1.17.2, skip-ui 1.57.0
- Fuse app; Kotlin 2.3.0, compileSdk 35; reproduces on emulator and devices

## Summary

`GeometryReader` (and the shared `onGloballyPositionedInRoot`/`InWindow` helpers, and `onGeometryChange`) write raw float `Rect` bounds into Compose state on every global-position callback, with no tolerance/epsilon. When the measured content can jitter at sub-pixel scale — or sizes itself against the proxy — this closes a write→recompose→remeasure→write loop that recomposes **continuously while the screen is fully idle**.

## Mechanism

`GeometryReader.Render` remembers a `Rect?` and gates its content on reading it back:

```swift
// Layout/GeometryReader.swift:26–33
@Composable override func Render(context: ComposeContext) {
    let rememberedGlobalFramePx = remember { mutableStateOf<Rect?>(nil) }
    Box(modifier: context.modifier.fillSize().onGloballyPositionedInRoot {
        rememberedGlobalFramePx.value = $0
    }) {
        if let globalFramePx = rememberedGlobalFramePx.value { ... content(proxy)... }
```

The only filter on the write is `!= Rect.Zero` — no comparison against the previous value beyond `mutableStateOf`'s bit-exact structural equality:

```swift
// Compose/ComposeExtensions.swift:130–137
@Composable func onGloballyPositionedInRoot(perform: (Rect) -> Void) -> Modifier {
    return self.onGloballyPositioned {
        let bounds = $0.boundsInRoot()
        if bounds != Rect.Zero { perform(bounds) }
    }
}
```

Any sub-pixel float variation between layout passes (remeasure, scroll settle, animations, or content whose size depends on the proxy value) produces a not-bit-identical `Rect` → state write → recomposition → new layout pass → new `Rect`. The sibling `onGeometryChangeErased` guards its *action* with an exact `!=` (AdditionalViewModifiers.swift:1002–1022) but still updates its internal state unconditionally — marginally better, same missing epsilon. `PresentationRoot` uses the same unguarded write for presentation bounds (Containers/PresentationRoot.swift:56–57).

## Repro

We hit this in our production Fuse app: a `GeometryReader` wrapping an empty-state layout centered against the reported height recomposed in a tight loop at idle — CPU measurably above zero with no interaction, animations stuttering — and scoping the `GeometryReader` down to the smallest subtree removed the trigger (which limits blast radius but not the underlying unguarded write). iOS with identical code is fully quiescent.

A standalone packaging of the same pattern with per-second body-evaluation counters is scene 2 (`GeometryLoopScene`) of https://github.com/Aecasorg/skip-fuse-perf-repro (expected: counters climb at idle on Android, stop after layout settles on iOS).

## Suggested fix

In `onGloballyPositionedInRoot`/`InWindow`: remember the last-delivered `Rect` and skip the callback when all four edges are within ~0.5px (one device pixel), or pixel-snap (round to nearest px) before comparing/writing. Apply the same guard to `GeometryReader`'s remembered state and `onGeometryChangeErased`'s internal write. This matches iOS behavior, where geometry callbacks are coalesced and don't feed back at sub-pixel granularity.

## Offer

This looks like a small, self-contained change (one helper + two call sites). We'd be glad to submit a PR with the epsilon guard plus the repro — is this direction acceptable?

Aecasorg:
Fixes #488.

## Problem

`GeometryReader`, `onGeometryChange`, and `PresentationRoot` write raw float `Rect` bounds from global-position callbacks into remembered state that gates their composed content. Bounds can differ from the previous layout pass by sub-pixel amounts (float rounding, scroll settle, in-progress animations, or content that sizes itself against the reported geometry), and `mutableStateOf`'s structural equality only deduplicates bit-identical rects — so a sub-pixel delta closes a write → recompose → remeasure → write loop that recomposes continuously while the screen is idle. We hit this as a sustained idle-recomposition loop in our production Fuse app (skip-ui 1.57-era): a `GeometryReader`-centered empty state recomposed continuously at rest, and scoping the `GeometryReader` to a smaller subtree removed the trigger — which limits blast radius but leaves the unguarded write in place. Full mechanism citations in #488.

## Change

Adds `Rect.isApproximatelyEqual(to:)` (all four edges within `boundsEpsilonPx = 0.5f` — half a device pixel, below anything visually meaningful) and guards the three gating state writes with it:

- `GeometryReader.Render` — `rememberedGlobalFramePx`
- `onGeometryChangeErased` — `globalFramePx`
- `PresentationRoot` — `presentationBounds`

Deliberately **not** changed: the shared `onGloballyPositionedInRoot`/`InWindow` helpers still deliver every callback. Deduplicating there would silently change semantics for callers whose derived values depend on more than the bounds — e.g. the safe-area edge probing in `ComposeLayouts.swift` recomputes edges from bounds *plus* the current safe area, and must not be suppressed when the safe area changes while bounds don't. Instead the helpers' doc comment now warns that callers gating content on remembered bounds should guard their writes.

The comparison intentionally avoids `abs()` — the transpiled `skip.lib` form is not comparable against a Kotlin `Float` (caught by the Android Kotlin compile).

## Verification

Verified with CI-built APKs of a standalone counter-instrumented app (https://github.com/Aecasorg/skip-fuse-perf-repro), on an API 35 emulator, comparing stock skip-ui 1.57.0, stock 1.59.1, and 1.59.1 + this patch:

- **Regression-free:** four GeometryReader shapes (plain fillSize, inside ScrollView, refreshable ScrollView, proxy-derived text) plus unrelated-state and onChange control scenes behave identically with and without the patch — same body-evaluation counts, same settle points.
- **Honest disclosure:** the *pure idle loop* did **not** reproduce in those minimal synthetic shapes at either skip-ui version on the emulator — counters settle at 26–28 evaluations and stay flat. The production loop we observed evidently needs conditions the minimal scenes don't capture (device density, concurrent animation, richer layout). We're framing this change as a cheap, verified-safe guard against a demonstrated-in-production hazard rather than claiming a synthetic before/after.

Happy to adjust the epsilon, naming, or approach — including pushing the guard into the shared helpers instead if you'd rather own the call-site audit.

### DIFF
--- a/Sources/SkipUI/SkipUI/Compose/ComposeExtensions.swift
+++ b/Sources/SkipUI/SkipUI/Compose/ComposeExtensions.swift
@@ -127,6 +127,11 @@ extension Modifier {
     }
 
     /// Invoke the given closure with the modified view's root bounds.
+    ///
+    /// - Warning: Callbacks can deliver bounds that differ from the previous pass by sub-pixel
+    ///   amounts (float rounding, scroll settle, in-progress animations). Callers that gate
+    ///   composed content on remembered bounds state should guard their state write with
+    ///   `Rect.isApproximatelyEqual(to:)` to avoid an idle recomposition loop.
     @Composable func onGloballyPositionedInRoot(perform: (Rect) -> Void) -> Modifier {
         return self.onGloballyPositioned {
             let bounds = $0.boundsInRoot()
@@ -182,6 +187,30 @@ extension PaddingValues {
             end: (insets1.trailing + insets2.trailing).dp,
             bottom: (insets1.bottom + insets2.bottom).dp
         )
+    }
+}
+
+/// The maximum per-edge delta at which two bounds rects are treated as visually identical.
+///
+/// Global-position callbacks can report bounds that differ from the previous layout pass by
+/// sub-pixel amounts (float rounding, scroll settle, in-progress animations). Writing such a
+/// rect into gating state closes a write → recompose → remeasure → write loop that recomposes
+/// continuously while the screen is idle. Half a pixel is below anything visually meaningful.
+let boundsEpsilonPx = Float(0.5)
+
+extension Rect {
+    /// Whether every edge of this rect is within `boundsEpsilonPx` of `other`'s.
+    ///
+    /// Use to guard remembered-bounds state writes in global-position callbacks; see
+    /// `onGloballyPositionedInRoot`.
+    func isApproximatelyEqual(to other: Rect?) -> Bool {
+        guard let other else {
+            return false
+        }
+        return abs(left - other.left) < boundsEpsilonPx
+            && abs(top - other.top) < boundsEpsilonPx
+            && abs(right - other.right) < boundsEpsilonPx
+            && abs(bottom - other.bottom) < boundsEpsilonPx
     }
 }
 
--- a/Sources/SkipUI/SkipUI/Compose/ComposeExtensions.swift
+++ b/Sources/SkipUI/SkipUI/Compose/ComposeExtensions.swift
@@ -207,11 +207,16 @@ extension Rect {
         guard let other else {
             return false
         }
-        return abs(left - other.left) < boundsEpsilonPx
-            && abs(top - other.top) < boundsEpsilonPx
-            && abs(right - other.right) < boundsEpsilonPx
-            && abs(bottom - other.bottom) < boundsEpsilonPx
+        return isWithinEpsilon(left, other.left)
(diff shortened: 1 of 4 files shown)


Aecasorg:
recheck

Issue #487, opened 2026-07-24
Presentation open path: open-sheet content re-Evaluates on every presenter recomposition, a guaranteed empty first frame, a preference settle pass, and a new Dialog window per presentation


Aecasorg:
## Environment

- skip 1.9.4, skip-fuse-ui 1.17.2, skip-ui 1.57.0
- Fuse app; Kotlin 2.3.0, compileSdk 35, Material3 `ModalBottomSheet`

## Summary

Opening and holding a sheet/alert/cover pays five separable framework costs that together make presentations feel noticeably less fluid than iOS running the identical SwiftUI code. Filing as one issue because they share the `Presentation.swift` open path — but each is independently fixable, and any subset helps.

## a) Open sheet content re-`Evaluate`s on every presenter recomposition

`.sheet`/`.alert`/`.fullScreenCover` install a `PresentationModifier` whose action runs inline in the **presenter's** render scope (View/ModifiedContent.swift:69–71 `action?(context)`; Layout/Presentation.swift:1286–1290). While presented, every presenter recomposition re-evaluates the sheet's entire content graph:

```swift
// Layout/Presentation.swift:100–103
if isPresentedValue || sheetState.isVisible {
    ...
    let contentRenderables = ComposeBuilder.from(content).Evaluate(context: context,
        options: EvaluateOptions(isKeepNonModified: true).value)
```

A busy presenter (timers, observable ticks, geometry writes) rebuilds the open sheet's full `Renderable` graph each frame even though the sheet draws in a separate window.

**Suggested fix:** cache `contentRenderables` keyed on content identity (`remember`/`derivedStateOf`), or move the `Evaluate` inside the `ModalBottomSheet` scope; the up-front evaluation is only needed to harvest modifiers — memoize that harvest.

## b) Guaranteed empty first frame in `PresentationRoot`

Bounds seed to `Rect.Zero`, real bounds arrive via a post-layout callback, and content is gated on non-Zero — so the first composition of every presentation renders nothing:

```swift
// Containers/PresentationRoot.swift:40, 56–60
let presentationBounds = remember { mutableStateOf(Rect.Zero) }
...
.onGloballyPositionedInWindow { presentationBounds.value = $0 }
...
guard presentationBounds.value != Rect.Zero else { return }
```

**Suggested fix:** read insets without the layout round-trip (`WindowInsets.safeDrawing` composition locals) so the first frame renders content.

## c) Presentation preferences need an extra composition to settle

[...]

Issue #486, opened 2026-07-24
One @Observable read re-runs the whole container's Render scope, and lazy items re-Evaluate every pass — stable ids never translate into Compose skipping


Aecasorg:
## Environment

- skip 1.9.4, skip-fuse-ui 1.17.2, skip-ui 1.57.0, skip-model 1.7.5
- Fuse app; Kotlin 2.3.0, compileSdk 35
- Related: #462 (NavigationStack recomposition) looks like a symptom of the same granularity design; #478 touched the `animateItem` corner of this from another angle.

## Summary

Two related design points in SkipUI combine so that state invalidation on Android is per-*container* where SwiftUI (and idiomatic Compose) is per-*view* / per-*row*:

1. Observation registers on the enclosing container's single restart scope, so one `@Observable` write re-runs the whole container and all its children.
2. Lazy containers rebuild their item factories and re-evaluate all visible items on every pass, so even perfectly stable `ForEach` ids don't let Compose skip unchanged rows.

Together these mean the cost of any state change scales with everything on screen rather than with what changed. App code can shrink the blast radius (small child views, reads pushed to leaves, lazy containers) but can never reach per-view granularity, because both mechanisms are framework-owned.

## Mechanism A — container-scope invalidation

`ComposeBuilder` content lambdas return a `ComposeResult`, which deliberately makes them non-restartable:

```swift
// Compose/ComposeBuilder.swift:49
// - Note: Returning a result from `content` is important. This prevents Compose from recomposing
//   `content` on its own. Instead, a change that would recompose `content` elevates to our void
//   `Renderable.Render`.
```

`View.Evaluate` (View/View.swift:67) is a value-returning `@Composable` where each child's `body` actually executes, and containers call it inline as the first line of their own restartable `Render`:

```swift
// Containers/VStack.swift:53
let renderables = content.Evaluate(context: context, options: 0).filter { !$0.isSwiftUIEmptyView }
```

So every `@Observable` read in *any* child body registers on the container's single restart scope; a mutation bumps the backing state (skip-model MutableStateBacking.swift:46, `state[index].value += 1`) and invalidates that one coarse scope, re-running the container and re-evaluating every child — including children that never read the mutated value. `StateTracking.pushBody/popBody` manage the read cursor but do not open a per-view restart group.

## Mechanism B — lazy items re-evaluated every pass, no memo boundary

[...]

PR #505, opened 2026-08-06
Compose sheets with final presentation-preference values on first composition; stop presenter churn re-evaluating confirmation dialogs

Aecasorg:
Addresses the preference-settle and presenter-churn items — (c) and part of (a) — of #487.

## What changes

**1. Sheets compose with final presentation-preference values on their first composition.**

`preference(key:value:)` now emits `PreferenceModifier`, a `SideEffectModifier` subclass that carries its key and value as inspectable fields (contribution behavior is byte-for-byte unchanged). `SheetPresentation` harvests statically-applied `presentationDetents` / `presentationDragIndicator` / `interactiveDismissDisabled` values from the content renderables it already evaluates (the same `forEachModifier` scan as the existing `backDismissDisabled` special case) and constructs the preference states with the harvested values as `initialValue`.

Result: a sheet with `.presentationDetents([.medium])` composes at medium-detent geometry on frame 1 instead of full height with a visible jump one composition later — and because the later collector contribution equals the seed, the settle recomposition itself disappears. Dynamic preference updates flow through the collectors exactly as before; preferences applied deeper than the content's own modifier chain keep the previous settle behavior (no regression, harvest is additive).

**2. `ModalBottomSheet` arguments stabilized.** The shape is remembered keyed on its inputs, the dismiss callback is remembered over `rememberUpdatedState`, and `backDismissDisabled` is harvested once per presentation (only its first composition is honored anyway, per the existing comment). Fresh unequal instances per presenter pass were forcing the sheet machinery to recompose on every presenter recomposition.

**3. Confirmation dialogs stop re-running action/message bodies on presenter churn.** `ConfirmationDialogPresentation` fully evaluated its actions and message (`options: 0` — bodies run) in the presenter's composition scope on every presenter recomposition while open. The evaluation now runs inside the `ModalBottomSheet` content lambda — the dialog window's own composition.

## Verification

> **Superseded, 2026-08-13.** The A/B table immediately below is confounded: the "stock" side was built
> on skip-ui 1.57.0 / skip 1.9.4 / skip-fuse-ui 1.17.2 while the patched side was 1.59.1 / 1.9.5 / 1.18.1,
> so it cannot separate this patch from the version delta. It is left in place rather than deleted,
> because it has already been read. The corrected single-variable A/B — same toolchain on both sides,
> only skip-ui swapped, 10 open/close cycles per build — is in the comment below, and its honest summary
> is: the pre-settle observations go 10 → 0 and sheet-content evaluations 41 → 32, with presenter ticks
> unchanged at 44 on both sides as a control, and **no measurable smoothness win**. Independent later work
> on our own app supports that reading: the dominant cost on a slow sheet turned out to be the body's
> lazy-container shape (see #508), not preference settling.

On-device A/B (API 35 emulator; standalone instrumented app https://github.com/Aecasorg/skip-fuse-perf-repro, scene 4: a sheet with `.presentationDetents([.medium])` presented over a presenter that ticks an unrelated `@Observable` every second):

| | stock 1.59.1 | patched |
|---|---|---|
| height seen by the sheet content's 1st evaluation | 801.5 (full height) → 438.1 ~50ms later, visible jump | **438.1 from the 1st evaluation** |
| sheet-content evaluations during open | n | n − 1 (settle pass gone) |
| evaluations per presenter tick while open | +1 | +1 (unchanged — see note) |

Note on the last row: content re-evaluation on presenter churn persists because the content lambda's captures (fresh renderables per presenter pass) invalidate the dialog composition — that residue is the bridged-peer identity problem (skip-bridge#113 / #486 territory) and is out of scope here. We considered memoizing `contentRenderables` as the issue suggested, but with `isKeepNonModified` evaluation the sheet's liveness for builder-level state rides on the presenter-scope re-evaluate — caching it without content identity would freeze open sheets, so we deliberately did not.

Alert presentation has the same presenter-scope evaluation pattern (plus text-field state extraction) — left for a follow-up to keep this reviewable.

Independent of #499 (different files except `Presentation.swift`, no overlapping hunks); the two merge in either order.

Happy to adjust naming, split the commits into separate PRs, or extend the harvest to alerts here if you'd prefer.

### DIFF
--- a/Sources/SkipUI/SkipUI/Environment/PreferenceKey.swift
+++ b/Sources/SkipUI/SkipUI/Environment/PreferenceKey.swift
@@ -181,16 +181,35 @@ struct PreferenceNode<Value>: Equatable {
         return lhs.value == rhs.value
     }
 }
+
+/// The modifier produced by `preference(key:value:)`.
+///
+/// Contributes through the standard collector at render time like any side effect, but
+/// carries its key and value as inspectable fields so that presentation containers can
+/// harvest statically-applied preferences from an evaluated renderable chain *before*
+/// their first composition (see `SheetPresentation`), instead of waiting for the
+/// contribution to settle a composition later.
+final class PreferenceModifier: SideEffectModifier {
+    let key: Any
+    let value: Any?
+
+    init(key: Any, value: Any?) {
+        self.key = key
+        self.value = value
+        super.init()
+        self.action = { context in
+            PreferenceValues.shared.contribute(context: context, key: key, value: value)
+            return ComposeResult.ok
+        }
+    }
+}
 #endif
 
 extension View {
     // SKIP @bridge
     public func preference(key: Any, value: Any?) -> any View {
         #if SKIP
-        return ModifiedContent(content: self, modifier: SideEffectModifier { context in
-            PreferenceValues.shared.contribute(context: context, key: key, value: value)
-            return ComposeResult.ok
-        })
+        return ModifiedContent(content: self, modifier: PreferenceModifier(key: key, value: value))
         #else
         return self
         #endif
--- a/Sources/SkipUI/SkipUI/Layout/Presentation.swift
+++ b/Sources/SkipUI/SkipUI/Layout/Presentation.swift
@@ -92,15 +92,17 @@ private let AlertDialogMaxWidth: Dp = 560.dp
 
 // SKIP INSERT: @OptIn(ExperimentalMaterial3Api::class)
 @Composable func SheetPresentation(isPresented: Binding<Bool>, isFullScreen: Bool, context: ComposeContext, content: () -> any View, onDismiss: (() -> Void)?) {
-    let interactiveDismissDisabledPreference = rememberSaveable(stateSaver: context.stateSaver as! Saver<Preference<Bool>, Any>) { mutableStateOf(Preference<Bool>(key: InteractiveDismissDisabledPreferenceKey.self)) }
-    let interactiveDismissDisabledCollector = PreferenceCollector<Bool>(key: InteractiveDismissDisabledPreferenceKey.self, state: interactiveDismissDisabledPreference)
-
     let sheetState = rememberModalBottomSheetState(skipPartiallyExpanded: true)
     let isPresentedValue = isPresented.get()
     if isPresentedValue || sheetState.isVisible {
         // Don't fully evaluate content until we set up the presented environment. For now we just want
-        // to get at the modifiers to look for `BackDismissDisabled`
+        // to get at the modifiers to look for `BackDismissDisabled` and statically-applied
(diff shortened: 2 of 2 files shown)


Aecasorg:
Pushed a third commit: the alert dialog's `onDismissRequest` gets the same remembered-callback treatment as the sheet and confirmation-dialog sites in this PR — same file, same pattern, so it seemed to belong here rather than in a PR of its own.

Worth being explicit that this is a consistency fix only: it does **not** address `AlertPresentation`'s presenter-scope evaluation of actions/message. Unlike the two sites above, that evaluation cannot simply move inside the dialog, because `SkipAlertDialog`'s slot API needs the button structure (neutral-button count, whether a cancel button exists, whether there are text fields) *before* the `Dialog` call — and discovering that structure is the evaluation itself. Fixing it properly looks like reshaping `SkipAlertDialog` to take a builder that discovers structure inside the `Dialog`, which also runs through the text-field slot where re-keying would cost `FocusRequester`/`TextFieldValue` state. Happy to attempt it if you'd like it in scope, but it seemed wrong to bundle a change of that size in here.

Aecasorg:
Re-ran the verification properly, because the A/B in the PR description was confounded: its "stock" side was skip-ui 1.57.0 with skip 1.9.4 / skip-fuse-ui 1.17.2, while the patched side was built on 1.59.1 with skip 1.9.5 / skip-fuse-ui 1.18.1 — so it could not distinguish this change from the upstream version delta. Apologies; that was sloppy of me.

Redone with skip-ui as the only variable (skip 1.9.5 / skip-fuse-ui 1.18.1 both sides, upstream `1.59.1` vs the same tree plus these commits), API 35 emulator, 10 scripted sheet open/close cycles per build, sheet declaring `.presentationDetents([.medium])` over a presenter ticking an unrelated `@Observable` every second:

| Metric over 10 opens | upstream 1.59.1 | with these commits |
|---|---|---|
| sheet-content evaluations observing the **pre-settle full height** (801.5 rather than the 438.1 medium detent) | **10 — exactly one per open** | **0** |
| sheet-content body evaluations | 41 | 32 |
| geometry-proxy evaluations | 346 | 326 |
| presenter ticks (control variable) | 44 | 44 |

So the settle pass is deterministic upstream — one per sheet open, every time — and it is gone with the preference seeding, along with ~22% of the sheet-content evaluations for the same number of presenter ticks.

Being straight about what this does **not** show: frame-level timing over the same runs was 7.8% vs 8.7% janky frames and 32ms vs 34ms at the 95th percentile — i.e. no measurable end-user smoothness win, and the small differences sit inside single-run emulator noise in both directions (the 99th percentile went the other way, 57ms vs 53ms). I would not claim a perceptible improvement from this data, and an emulator is a poor instrument for that question anyway. The defensible claim is the one above: a guaranteed extra composition per presentation, with its geometry jump, is removed.

Issue #517, opened 2026-08-31
Reorder drag has no visible affordance on dark backgrounds: dragging flag is spent on an invisible shadow


humbuckr:
## Summary

`List` computes whether a row is being dragged and spends it entirely on
`Modifier.shadow(8.dp)`. On a dark app that shadow is invisible, so a reorder has
**no visible affordance at all** — the row you are dragging looks identical to
every other row. There is no way to reach the flag from Swift to supply your own
treatment.

## Where

`Sources/SkipUI/SkipUI/Containers/List.swift`, `RenderReorderableItem`:

```swift
@Composable private func RenderReorderableItem(reorderableState: ReorderableLazyListState, key: String, modifier: Modifier, content: @Composable (Modifier) -> Void) {
    ReorderableItem(state: reorderableState, key: key, defaultDraggingModifier: modifier) { dragging in
        var itemModifier = Modifier.detectReorderAfterLongPress(reorderableState)
        if dragging {
            let elevation = animateDpAsState(8.dp)
            itemModifier = itemModifier.shadow(elevation.value)
        }
        content(itemModifier)
    }
}
```

`dragging` is known here and discarded. A shadow communicates elevation by
darkening what is *around* the row, which does nothing over `#000`/`#141414`.
Material's own answer on dark surfaces is tonal elevation — the surface gets
*lighter* — which a shadow cannot express.

## Why an app cannot work around it

- **The flag never reaches Swift.** `onMove` is invoked from `onDragEnd`
  (`rememberReorderableLazyListState(..., onDragEnd:)` → `commitMove()`), so
  SwiftUI's `.onMove` fires on drop. During the drag the app knows nothing.
- **A gesture of one's own does not fire.** `Modifier.detectReorderAfterLongPress`
  claims the long press first, so `.onLongPressGesture(..., onPressingChanged:)`
  on the row never sees the press. Verified on device.
- **And that callback is one-way regardless.** `onLongPressChange()` calls
  `onChanged(true)`; nothing ever calls it with `false`
  (`System/Gesture.swift`), so a press-derived highlight would latch on rather
  than follow the finger.

## Suggested directions

1. Apply a tonal treatment as well as the shadow when dragging — e.g. compose the
   row over a surface tinted by elevation, so dark themes get feedback that a
   shadow cannot give them.
2. Or make it configurable, in the shape `material3Ripple` already uses: an
   environment hook letting the app supply the dragging modifier.
[...]

PR #518, opened 2026-08-31
List: make the reorder drag visible on dark backgrounds

humbuckr:
Fixes the problem reported in #517.

## The problem

`RenderReorderableItem` styles the dragged row with `Modifier.shadow(8.dp)` and
nothing else. A shadow expresses elevation by darkening what is *around* the
row — so over a dark page it expresses nothing. On `#000` there is nothing left
to darken, and the row under your finger is pixel-identical to the rows at rest.
A reorder on a dark-themed app has no visible affordance at all.

## The change

Draw a tonal lift alongside the shadow, which is Material's own answer for
elevation on dark surfaces: the surface itself gets lighter.

```swift
let dragTint = MaterialTheme.colorScheme.onSurface.copy(alpha: Float(0.10))
itemModifier = itemModifier.drawWithContent {
    drawContent()
    drawRect(color: dragTint)
}
```

Two choices worth flagging for review:

- **Over the content, not behind it.** A list row almost always paints its own
  background, so a tint drawn behind it would never be seen.
- **`onSurface`, not a fixed colour**, so it follows the scheme in both
  directions — lighter on dark themes, darker on light ones. Light themes keep
  the shadow they already had and gain a subtle press-like tint; dark themes gain
  the only feedback they can receive.

Happy to put it behind an environment hook in the shape of `material3Ripple`
instead, if you would rather this be opt-in than a default.

## Why apps cannot work around it

- `onMove` is invoked from `onDragEnd`, so SwiftUI code learns about the move
  only on drop.
- `Modifier.detectReorderAfterLongPress` claims the long press, so an app's own
  `.onLongPressGesture(..., onPressingChanged:)` on the row never fires.
- And that callback is one-way regardless: `onLongPressChange()` calls
  `onChanged(true)` and nothing ever calls it with `false`.

## Testing

Built and run on device: Android 14, Samsung SM-A326U (Dimensity 700), release
build, a 24-row playlist on a `#121212` page with `#252525` rows. Before the
patch, dragging produced no visible change; after it, the dragged row lifts while
held and returns to rest on drop. Light-theme rendering unchanged apart from the
new tint.

### DIFF
--- a/Sources/SkipUI/SkipUI/Containers/List.swift
+++ b/Sources/SkipUI/SkipUI/Containers/List.swift
@@ -59,6 +59,7 @@ import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.draw.clipToBounds
+import androidx.compose.ui.draw.drawWithContent
 import androidx.compose.ui.draw.shadow
 import androidx.compose.ui.layout.onSizeChanged
 import androidx.compose.ui.geometry.CornerRadius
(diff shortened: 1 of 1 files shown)


marcprux:
This looks useful?

Could you post a before and after screenshot of a row being dragged with this effect?

humbuckr:
Hey Marc - I am new to Git  and skip-ui.  i dont have a screenshot from
before now.  sorry.

Issue #263, opened 2025-10-15
Custom ButtonStyle not supported; `ButtonStyle` declaration incorrect


dfabulich:
https://developer.apple.com/documentation/SwiftUI/ButtonStyle
https://www.avanderlee.com/swiftui/swiftui-button-styles/

`ButtonStyle` is a protocol, allowing you to implement its `makeBody(configuration:)` function to implement a custom button style.

SkipUI defines a struct called `ButtonStyle` with no `makeBody` implementation; it's basically an `enum`.

https://github.com/skiptools/skip-ui/blob/a72881a367d0d6a663796949a09d9666b8efde07/Sources/SkipUI/SkipUI/Controls/Button.swift#L231-L245

As a result, if you try to define a custom `ButtonStyle` with a `makeBody` implementation, Skip will spew a bunch of confusing errors ("Skip is unable to determine the owning type for member"); if you do manage to convince it it compile, it simply won't work.

PR #523, opened 2026-09-12
Support ButtonStyle and PrimitiveButtonStyle protocols

gurmeherchawla:
Replaces the raw-value `ButtonStyle` bridging struct with proper `ButtonStyle` and `PrimitiveButtonStyle` protocols matching SwiftUI's API, enabling custom button styles via `makeBody` on Android.

Labels: enhancement

### What changed

**`ButtonStyle` / `PrimitiveButtonStyle` protocols**
The old `ButtonStyle` struct was a raw-value enum used purely for bridging. It is replaced by two protocols:
- `PrimitiveButtonStyle` — for styles that control their own interaction (matches SwiftUI)
- `ButtonStyle` — for styles that receive `isPressed` state and defer interaction to the framework

All built-in styles (`DefaultButtonStyle`, `PlainButtonStyle`, `BorderedButtonStyle`, `BorderedProminentButtonStyle`, `BorderlessButtonStyle`, `M3TextButtonStyle`) are now concrete structs conforming to `PrimitiveButtonStyle`. They continue to render natively in Compose without going through `makeBody`.

**`StackedButtonStyle`**
Introduced to track the active style chain in the environment. When `Button(configuration)` is called inside a custom `makeBody`, it renders in the next outer style rather than recursing — matching SwiftUI's behaviour. The `source` reference prevents a modifier from making itself its own parent when it re-evaluates within its own scope.

**`RenderCustomStyleButton`**
New rendering path for non-built-in styles. Calls `makeBody`, sets `_buttonStyle` to the parent stack node, then composes the body so that any nested `Button(configuration)` delegates outward correctly.

**Bridge support**
`BridgedButtonStyle` and `BridgedPrimitiveButtonStyle` allow natively-compiled styles (from skip-fuse-ui) to cross the bridge. `ButtonStyleBridgedConfiguration` exposes `environmentSupport(forKey:)` so the bridge layer can sync `@Environment` values declared on the style at each button's Compose position. A new `buttonStyle(isPrimitive:environmentKeys:bridgedMakeBody:)` overload is the bridge entry point.

**Other fixes**
- `ToggleStyle` static properties (`automatic`, `button`, `switch`) were incorrectly typed as `ButtonStyle` — fixed.
- `Picker` and `NavigationLink` list-item rendering updated to use the new `isListItemStyle` helper.

### Related PR

skip-fuse-ui: https://github.com/skiptools/skip-fuse-ui/pull/143


gurmeherchawla:
closes #263

gurmeherchawla:
@marcprux

PR #524, opened 2026-09-13
Honour .disabled on menu items

### LINKED ISSUE #246: Menu does not respect disabled state for Button
On iOS, the third button will be disabled and grayed out, but on Android it is not disabled:

```swift
Menu("Default") {
    Button("Option 1") { logger.log("Option 1") }
    Button("Option 2") { logger.log("Option 2") }
    Button("Option 3") { logger.log("Option 3") }.disabled(true)
}
```

Ceylo:
## What

Fixes #246. A menu item is rendered from its stripped `Button`, so the `DisabledModifier` in its chain was dropped and `DropdownMenuItem` always got `enabled = true`: a `.disabled(true)` item stayed tappable and ran its action.

The item's modifier chain is now read and passed as `enabled`. SkipUI gives the item an explicit tint, which overrides `DropdownMenuItem`'s own disabled colours, so a disabled item's tint is dimmed with `ContentAlpha.disabled`, the same idiom `Button` uses. A disabled nested `Menu` opens as a copy of its content wrapped in `DisabledModifier(true)`. That matches iOS 26.5, which opens it with all its items disabled.

## Before / after

Showcase → Menu → the new `.disabled(true) items: N` menu (skiptools/skipapp-showcase#122). Options 2, 3, 5 and the nested Option 4 are disabled. Each of them increments the counter when its action runs.

| | before | after | iOS 26.5 |
|---|---|---|---|
| menu open |  |  |  |
| after tapping every disabled item |  |  |  |
| disabled nested menu open | — (runs its items) |  |  |

Android shots are Fuse. Counter after the taps: **4 → 0**. Text colour of Options 2, 3 and 5, read from the screenshot pixels: **rgb(48, 50, 59) → rgb(165, 166, 174)**. Option 1 and Nested stay at rgb(48, 50, 59). Lite reads the same values.

## Side effects

- **Enabled items are unchanged.** The new code only runs for a disabled item. The Default menu and the Context Menu playground (which has a destructive item) are pixel-identical before and after, except the status bar. These were Lite shots, taken on an earlier revision of this branch that already had the `enabled` change. The later dimming and nested-menu changes never touch an enabled item.

  | Context Menu, before | after |
  |---|---|
  |  |  |
- **`ContextMenu`** renders its items through the same path, so it gets the fix too.
- **A disabled `Section`** needed no change. Its content is flattened when evaluated and each child carries the modifier. The test covers it.
- **Not fixed here:** menu-style `Picker` options still ignore `.disabled`, because each option is rewrapped with only a `TagModifier`. This was already the case and is left for a follow-up.

## Verified

- `swift test`: 95 Swift tests, 36 skipped, 0 failures. JUnit: 94 tests, 89 passed, 5 skipped, 0 failed. Adds `testMenuDisabledItems` and `testMenuEnabledItem` next to `testMenuAccessibilityIdentifier`.
- Mutation runs, each failing only on its own assertion:

  | Menu.swift | result |
  |---|---|
  | as on `main` | `testMenuDisabledItems` fails `(is not enabled)` on `menu.item.disabled` |
  | nested-menu change removed | `testMenuDisabledItems` fails `(is not enabled)` on `menu.item.nested` |
  | nested-menu condition flipped | `testMenuEnabledItem` fails `(is enabled)`, `testMenuDisabledItems` fails `(is not enabled)` |
  | this PR | both pass |
- Showcase built through Gradle with this branch as its skip-ui (`SKIP_DEPENDENCY_ROOT`), then run in Lite and Fuse on an API 37 Android emulator. Also run on an iPhone 17 simulator (iOS 26.5) as the reference.

---

Skip Pull Request Checklist:

- [X] REQUIRED: I have checked whether this change requires a corresponding update in the [Skip Fuse UI](https://github.com/skiptools/skip-fuse-ui) repository (link related PR if applicable). No API change, so no paired PR.
- [X] OPTIONAL: I have added an example of any UI changes to the [Showcase](https://github.com/skiptools/skipapp-showcase) sample app: skiptools/skipapp-showcase#122

-----

- [X] AI was used to generate or assist with generating this PR. *Please specify below how you used AI to help you, and what steps you have taken to manually verify the changes*.

Written with Claude Code: the patch, the tests and the playground. Every claim above comes from a run: Robolectric tests with the mutation runs above, and the Showcase on an emulator (Lite and Fuse) and a simulator, with before/after screenshots and pixel colour readings.


marcprux:
Great improvement, thanks!

Issue #462, opened 2026-06-16
NavigationStack recomposing problem


fwilldev:
Hello,

I am currently migrating my app to skip.

With the help of Claude Code I made a lot of progress, but I hit a blocker due to infinite recomposes - which makes my app unbearably laggy.

I am aware of https://github.com/skiptools/skip-ui/pull/441 and https://github.com/skiptools/skip-ui/pull/440 - so it seems no fix is currently available for this.

**Environment**
  - Skip Fuse app; skip-ui 1.56.0, skip-fuse-ui 1.17.0, skip-model 1.7.5, skip 1.9.3
  - Physical device: Samsung Galaxy A16, Android 16 (API 36)

 **Evidence (Layout Inspector, app sitting idle)**
  - NavDisplay, RenderEntry, and collectPreferences (inside the nav entry) increment every frame - continuously, with no user interaction.
  - A bare NavigationStack { Text("x") } is idle (0 recompositions). The loop only appears once the nav entry's content contributes preferences (real screens with a nav bar/normal content).

 **What I already tried (none fixed it):**
 - Removed `SideEffect { saveableStateHolder.removeState(true) }` from Main.kt — reduced per-frame cost but the loop continued.

So because this is a big roadblocker for me, my questions are:

1. Is there any app-side workaround to avoid the per-frame preference recomposition i am not aware of, until a fix lands?

Happy to share more info if needed.

marcprux:
It is possible that https://github.com/skiptools/skip-ui/pull/441 might fix it. Can you try running your app locally with that version (either by checking out https://github.com/skiptools/skip-ui/pull/441 locally and adding it to your shared Xcode workspace, or by just pinning your skip-ui dependency to the fork & branch for that PR), and see if it fixes the recompose issue?

fwilldev:
I did that and it didn't fix the problem....

Also tried to debug even further, but I am struggeling to find the real root cause of the rerenderings. Could be the NavigationStack, but I am not that sure anymore - because a simple Repro with just some NavigationStacks dont reproduce these recompositions, with various versions of SkipUI, etc. 

My SwiftUI App is pretty complex with lots of multi-child screens, an AuthGuard, a SubscriptionManager (with RevenueCat, with Skips Implementation of RevenueCat).

Maybe someone can give me some hints where to look at but I know its pretty difficult todo that from the distance without the source-code.

I will pause the migration and will get back to it sometime in the future.

Sorry I cant contribute more to this issue.

Many thanks!

[...]

PR #528, opened 2026-09-18
Add a regression test that toolbar content keeps tracking state changes (#462)

vincentborko:
A regression test for #462: toolbar content must keep tracking state changes.

`ToolbarContentPreferences.==` compares the content views (Toolbar.swift). A producer that rebuilds its content on every composition never compares equal, so `PreferenceCollector.contribute` writes on every composition and `RenderEntry` reads that state in the same restart scope — the write invalidates the scope that produced it. Measured at rest in a Fuse app on an Android 15 emulator: 58.2 body evaluations/s on a screen with a toolbar, 0 without one, 0 with the comparison forced to report equality.

Forcing equality is the tempting fix, and this test says what it costs. It is not a fix itself; it fences off the wrong one so the real fix can be chosen on its merits.

## Verification

`swift test` on the transpiled side, this branch:

```
JUNIT TEST PASSED ToolbarTests.testToolbarContentTracksStateChanges (0.173)
JUNIT SUITES 10 TESTS 93 PASSED 88 (95.0%) FAILED 0 SKIPPED 5
```

Same run with `==` returning true for any two non-nil contents:

```
JUNIT TEST FAILED ToolbarTests.testToolbarContentTracksStateChanges (0.135)
  at skip.ui.ToolbarTests.testToolbarContentTracksStateChanges$SkipUI(ToolbarTests.kt:46)
JUNIT SUITES 10 TESTS 93 PASSED 87 (94.0%) FAILED 1 SKIPPED 5
```

Exactly one test moves; the other 92 are unchanged. The failure is `item-beta` not displayed: Compose does not store the new content, and the bar keeps rendering the views it collected first.

## Notes

The probe uses `ComposeView` so `.toolbar` receives a freshly built item per composition, which is the shape a bridged view tree produces. A transpiled `body` builds its content once and contributes the same instance, so it does not enter the loop.

Robolectric-only: `createComposeRule()` is not available on the host side, so the test throws `XCTSkip` there.

I could not reproduce the recomposition cycle itself under `createComposeRule()` — the entry content composed once and was skipped afterwards, so the cycle never closed. That gap is why this is a guard and not a fix.

One correction to my earlier comment on #462, measured afterwards: across 15 screens, 6 of 10 with a toolbar redraw at rest and 4 do not, while 2 of 5 without one do. A toolbar is therefore neither necessary nor sufficient for redraw at rest, and at least one other source exists that I have not identified.


vincentborko:
Fair question to raise myself: this PR fixes nothing, and a test-only PR is weaker than a test that ships with its fix. Say the word and I close it, or hold it until the fix lands and squash them together.

The case for taking it now is narrow. The test does not guard against a regression in current behaviour — it passes on `main` either way. It guards against one specific wrong fix for #462, weakening `ToolbarContentPreferences.==`, which is the first thing anyone reaching for that bug will try, because it silences the recomposition immediately. What it costs is invisible until you look at the bar: Compose then keeps the content it collected first, and a toolbar item stops tracking state. That is the trade I wanted written down somewhere executable rather than in an issue comment.

If you would rather have it with the fix, here is where I am on that, so you can judge whether it is worth waiting for.

The mechanism is settled: the producer writes a preference during composition, and `RenderEntry` reads that same state in the scope that hosts the producer, so the write invalidates its own producer. What is not settled is the cure, and the three candidates are not equal:

- **At the comparison** — ruled out by this test.
- **At the collector** — does not apply. The value really does differ by its own `==`; the collector is not writing something equal.
- **At the read** — the structural candidate, and the one your own comment in `RenderEntry` already anticipates: *"We would like to only process toolbar content in our topBar/bottomBar Composables, but composing custom ToolbarContent multiple times … prevents it from updating properly on recompose."* The parent needs `topLeadingItems.size > 0` for `showTopBar`, which is what keeps the read up top.

A shape that satisfies both: keep the views in a holder that is not snapshot state, so reading them does not subscribe the reader, and keep observable only what is cheaply and stably comparable. `.toolbar(id:bridgedContent:)` already carries an id, which is a candidate key.

I have not built that, and I would not send it blind. I cannot reproduce the cycle under `createComposeRule()` — the entry content composes once and is then skipped, so the loop never closes, which is the same gap noted on #462. Without a reproduction I can verify against, a change at the read site is a guess with a plausible story attached, and this is your framework, not mine. Measuring it needs a probe for body evaluations per second in a bridged app, which is the piece I would build first.

So: guard now and fix later, guard only with the fix, or not at all — your call.

vincentborko:
Closing this myself — I should not have opened it, and I can say why with numbers rather than taste.

Of the last 20 merged PRs here, 18 touch no tests at all, and the two that do are a colour-parsing follow-up and your own transaction work. The typical shape is one file, between +3 and +80 lines, changing behaviour: "Fix …", "Support …", "Implement …". This PR is one file, +70 lines, no behaviour change. It does not look like what gets merged here, and asking you to decide that was me handing over a judgement I could have made by reading the repo.

The measurement behind it stands and stays on #462, which is where it belongs: forcing `ToolbarContentPreferences.==` to report equality stops the idle recomposition and costs the toolbar its updates, because Compose then keeps the content it collected first. Anyone reaching for that shortcut can reproduce it in two minutes.

The branch stays on my fork. If a fix at the read site ever works out, the test comes back with it, in the same PR, where it is evidence for a change instead of a change by itself.

